PR #10, opened 2021-01-07
feat: ZwavejsServer instance

### LINKED ISSUE #9: Migrate addAPItoExpress to instance
For better integration with Z-Wave JS 2 MQTT we need to be able to:
 - Tear down Z-Wave JS Server
 - Update the Z-Wave JS driver (optional, we could tear down and set up again)

robertsLando:
Fixes #9


robertsLando:
@balloob Reverted linting

cli/server.ts
```diff
@@ -73,9 +73,13 @@ const expectedConfig = ["_", "config", "mock-driver"];
   });
 
   driver.on("driver ready", () => {
```
balloob, line 75: Let's make this callback `async` so we can use `await` instead of promises.

config.json
```diff
@@ -0,0 +1,3 @@
+{
```
balloob, line 1: This file is not used?

lib/server.ts
```diff
@@ -1,12 +1,43 @@
...
+export class ZwavejsServer {
+  private server: HttpServer;
+  private wsServer: ws.Server;
+  private sockets: Clients;
+
+  constructor(private driver: Driver, private options: ZwavejsServerOptions) {}
+
+  async start() {
+    this.server = createServer()
+    this.wsServer = new ws.Server({ server: this.server });
```
balloob, line 26: On what port are we listening for incoming connections now? Just any path ? 

We should update the README.

cli/server.ts
robertsLando: @balloob AFAIF it's not good practise to use async functions in event listeners as the EE doesn't handle the rejections. Even if we would be 100% sure it doesn't reject

config.json
robertsLando: Ops created it for testing. Can I add it to ignore instead?

lib/server.ts
robertsLando: The port of the server, 3000

balloob: WebSockets are based on the HTTP spec, and so will always rely on full URLs. 

I just checked and with this approach, any path works. So it's backwards compatible but we should still update the README to clarify that they should just connect to `/`.

balloob: We should also update `cli/client.ts` to no longer use `/zjs`

robertsLando: ok doing it now

robertsLando: done

cli/server.ts
balloob: You can use try…catch. It behaves just like `.catch()` on promises.

```ts
try {
  await server.start()
  console.info("Server listening on port 3000");
} catch (err) {
  console.error("Unable to start server", err);
  await driver.destroy();
}
```

balloob (approved):

cli/server.ts
robertsLando: I know, in other repo i maintain some user told me that is better to avoid

```js
something.on('eventName', async function(){})
```

It's just a bad pratcise even if you are sure (by using a try catch for example) that function never throws. Anyway will change it no problem :)

robertsLando:
@balloob I have moved to async/awat approach. Once this is merged I will go on with eslint

PR #16, opened 2021-01-08
chore: node cli module

robertsLando:
This PR setup the package for npm publish and replace minimalist with yargs


src/bin/server.ts
```diff
@@ -12,6 +14,7 @@ interface Args {
 const expectedConfig = ['_', 'config', 'mock-driver'];
 
 (async () => {
+  // @ts-ignore
```
balloob, line 17: We add types and then have to ignore the only place where the type are used? :D

src/lib/server.ts
```diff
@@ -170,8 +172,11 @@ interface ZwavejsServerOptions {
 }
 
 export class ZwavejsServer {
+  // @ts-ignore
```
balloob, line 175: What are we ignoring? Also, instead of ignore, we should use "expect error" so if anything ever changes, we are notified.

tsconfig.json
```diff
@@ -1,7 +1,20 @@
 {
   "compilerOptions": {
+    "target": "es5",
```
balloob, line 3: We shouldn't target ES5. It's a node app and even the latest LTS of Node supports more.

package.json
```diff
@@ -1,11 +1,17 @@
...
+  "bin": {
+    "zwave-server": "dist/bin/server.js",
+    "zwave-client": "dist/bin/client.js"
+  },
   "scripts": {
     "lint": "eslint --ext .ts .",
-    "lint:fix": "eslint --fix --ext .ts ."
+    "lint:fix": "eslint --fix --ext .ts .",
+    "build": "tsc -p .",
+    "local": "npm i -g && zwave-server"
```
balloob, line 14: We should not install things globally. Also, the zwave server needs to have CLI args passed.

robertsLando: This was just for local testing :)

tsconfig.json
robertsLando: Let me know what target you prefer to use

src/lib/server.ts
robertsLando: This is temporary until I found how to fix the error typescript is showing

src/bin/server.ts
robertsLando: This is temporary until I found how to fix the error typescript is showing

package.json
robertsLando: With this PR the module gets installed correctly. A user will run `npm install @zwave-js/server` and then can start the server with `zwave-server /dev/ttyACM0` or the client with `zwave-client`

tsconfig.json
balloob: Node 12 uses V8 7.8 which was released in 2019. So we can at least target ES 2018.

package.json
balloob: Yeah that sounds good. But then let's drop the local global install. It's not needed.

robertsLando:
@balloob This is ready

package.json
```diff
@@ -1,22 +1,28 @@
...
+  "main": "dist/lib/index.js",
+  "bin": {
+    "zwave-server": "dist/bin/server.js",
+    "zwave-client": "dist/bin/client.js"
+  },
   "scripts": {
     "lint": "eslint --ext .ts .",
-    "lint:fix": "eslint --fix --ext .ts ."
+    "lint:fix": "eslint --fix --ext .ts .",
+    "build": "tsc -p ."
```
balloob, line 13: Just to make sure we always have the latest build included with release:

```suggestion
    "build": "tsc -p .",
    "prepublishOnly": "rm -rf dist && npm run build"
```

balloob (approved):

robertsLando:
@balloob Humm any ideas how to merge this? :laughing:

balloob:
I rebased it for you, see #22.

balloob:
Merged in #22

robertsLando:
@balloob Thanks! I'm flooded by notifications from zwavejs2mqtt

PR #19, opened 2021-01-08
Add node.set_value

balloob:
Add node.set_value command.

### DIFF
--- a/lib/incoming_message.ts
+++ b/lib/incoming_message.ts
@@ -1,6 +1,18 @@
+import { ValueID } from "zwave-js";
+
 interface IncomingCommandStartListening {
   messageId: string;
   command: "start_listening";
 }
 
-export type IncomingMessage = IncomingCommandStartListening;
+interface IncomingCommandNodeSetValue {
+  messageId: string;
+  command: "node.set_value";
+  nodeId: number;
+  valueId: ValueID;
+  value: unknown;
+}
+
+export type IncomingMessage =
+  | IncomingCommandStartListening
+  | IncomingCommandNodeSetValue;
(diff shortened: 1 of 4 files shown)


lib/incoming_message.ts
```diff
@@ -1,6 +1,18 @@
...
+
 interface IncomingCommandStartListening {
   messageId: string;
   command: "start_listening";
 }
 
-export type IncomingMessage = IncomingCommandStartListening;
+interface IncomingCommandNodeSetValue {
+  messageId: string;
+  command: "node.set_value";
```
robertsLando, line 10: Isn't it better to split this in `instance: 'node'` and `command: 'setValue`  instead?

robertsLando: Instances will be `controller|node|driver`

marcelveldt: or perhaps "target"

marcelveldt:
we'll suit with this for now as we're still in POC/alpha stage but if/when we add more commands I think it's better to have a common message pattern such as:

{ "command": "mycommand", "data": optional_data_object, "messageId": "optionalIdAsString" }

and 

{ "event": "some_event", "data": optional_data_object }

and 

{ "result": "mycommand", "data": object, "messageId": idFromEarlierCommand }


In this case the nodeId should be part of data, as is the valueId.

lib/incoming_message.ts
balloob: The downside of doing this is that it's more difficult to map the types.

Now types and the actual commands are keyed by `command`. If we split it, it will be keyed by `(instance, command)`

PR #20, opened 2021-01-10
chore: separate instance handlers

Michsior14:
This separates the handing of messages based on instance type to avoid growing `server.ts` code.
I was not sure about the naming/files structure and will be happy to improve it depending on needs.

Hope this idea will be helpful ;)

### DIFF
--- a/lib/node/incoming_message.ts
+++ b/lib/node/incoming_message.ts
@@ -1,0 +1,15 @@
+import { ValueID } from "zwave-js";
+import { IncomingCommandBase } from "../incoming_message_base";
+import { NodeCommand } from "./command";
+
+export interface IncomingCommandNodeBase extends IncomingCommandBase {
+  nodeId: number;
+}
+
+export interface IncomingCommandNodeSetValue extends IncomingCommandNodeBase {
+  command: NodeCommand.setValue;
+  valueId: ValueID;
+  value: unknown;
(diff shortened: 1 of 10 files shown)


lib/node/incoming_message.ts
```diff
@@ -0,0 +1,15 @@
+import { ValueID } from "zwave-js";
+import { IncomingCommandBase } from "../incoming_message_base";
+import { NodeCommand } from "./command";
+
+export interface IncomingCommandNodeBase extends IncomingCommandBase {
+  nodeId: number;
+}
+
+export interface IncomingCommandNodeSetValue extends IncomingCommandNodeBase {
+  command: NodeCommand.setValue;
```
balloob, line 10: I am not a fan of this. It just means we are declaring everything in other places and in the end it's unclear what is happening where.

balloob: It's an improvement, we can always move things around later. Let's just go with it .

balloob (approved):
This looks great. Thanks!

lib/node/incoming_message.ts
Michsior14: I've did it only to make sure the property in `NodeResultTypes` matches the command ;) It's easy to misspell it especially when the number of commands grow.

Issue #3, opened 2021-01-05
add support for controller commands


marcelveldt:
add logic to issue controller commands

**must-have commands:**

- [x] beginInclusion #21
- [x] stopInclusion #21
- [x] beginExclusion #21
- [x] stopExclusion #21
- [x] removeFailedNode #21
- [x] replaceFailedNode #21
- [ ] Managing associations

**nice to have commands:**

- [x] healNode #21
- [x] beginHealingNetwork #21
- [x] stopHealingNetwork #21
- [x] isFailedNode #21

https://zwave-js.github.io/node-zwave-js/#/api/controller?id=controller-methods

PR #21, opened 2021-01-11
Add controller commands

balloob:
Building on top of @Michsior14 excellent structure I was able to quickly add all required controller commands from #3


lib/controller/message_handler.ts
```diff
@@ -39,6 +39,22 @@ export class ControllerMessageHandler {
...
+      case ControllerCommand.stopHealingNetwork: {
+        const success = await driver.controller.stopHealingNetwork();
+        return { success };
+      }
+      case ControllerCommand.isFailedNode: {
+        const success = await driver.controller.isFailedNode(message.nodeId);
+        return { success };
+      }
       default:
         throw new UnknownCommandError(
```
Michsior14, line 59: To avoid usage of `@ts-expect-error` you could just destruct `message` before the `switch` and pass `command` directly to `UnknownCommandError`.

Michsior14 (approved):
LGTM

PR #23, opened 2021-01-11
feat: add non-essential node commands

### LINKED ISSUE #5: add support for node commands
Add support of sending commands to Z-wave nodes, most important the setValue command

**must have commands:**

- [x] setValue

**nice to have commands:**

- [ ] firmware update
- [ ] refreshInfo
- [ ] getDefinedValueIDs
- [ ] getValueMetadata

Michsior14:
Closes #5

### DIFF
--- a/src/lib/node/incoming_message.ts
+++ b/src/lib/node/incoming_message.ts
@@ -12,4 +12,36 @@ export interface IncomingCommandNodeSetValue extends IncomingCommandNodeBase {
   value: unknown;
 }
 
-export type IncomingMessageNode = IncomingCommandNodeSetValue;
+export interface IncomingCommandNodeRefreshInfo
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.refreshInfo;
+}
+
+export interface IncomingCommandNodeGetDefinedValueIDs
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.getDefinedValueIDs;
+}
+
+export interface IncomingCommandNodeGetValueMetadata
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.getValueMetadata;
+  valueId: ValueID;
+}
+export interface IncomingCommandNodeAbortFirmwareUpdate
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.abortFirmwareUpdate;
+}
+export interface IncomingCommandNodeBeginFirmwareUpdate
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.beginFirmwareUpdate;
+  buffer: Buffer;
+  target?: number;
+}
+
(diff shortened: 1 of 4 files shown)


src/lib/node/outgoing_message.ts
```diff
@@ -1,5 +1,11 @@
+import { TranslatedValueID, ValueMetadata } from "zwave-js";
 import { NodeCommand } from "./command";
 
 export interface NodeResultTypes {
   [NodeCommand.setValue]: { success: boolean };
+  [NodeCommand.refreshInfo]: {};
+  [NodeCommand.getDefinedValueIDs]: { valueIds: TranslatedValueID[] };
```
Michsior14, line 7: Not sure if we want to return object or directly `TranslatedValueID[]`.

balloob: Always an object.

src/lib/node/incoming_message.ts
```diff
@@ -12,4 +12,36 @@ export interface IncomingCommandNodeSetValue extends IncomingCommandNodeBase {
...
+  valueId: ValueID;
+}
+export interface IncomingCommandNodeAbortFirmwareUpdate
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.abortFirmwareUpdate;
+}
+export interface IncomingCommandNodeBeginFirmwareUpdate
+  extends IncomingCommandNodeBase {
+  command: NodeCommand.beginFirmwareUpdate;
+  buffer: Buffer;
```
balloob, line 37: Our message format is limited to what can be represented in JSON.

balloob: Sending files as b64 encoded strings over web socket seems a bit weird too. How large are these generally ?

Michsior14: Right, while Buffer can be represented in JSON in JS, I have no idea about Python. Regarding the size I guess we need @marcelveldt advice ;)

balloob: Let's leave this command out for now so we can move forward with this PR. We can then do a separate discussion about this. 

btw, if you have some extra time, could use some help with reviewing my PRs :)

balloob (approved):

src/lib/node/incoming_message.ts
marcelveldt: firmware update is pretty advanced so I'd say leave it out for now. Users can fire up zwave2mqtt control panel to do firmware updates so there's a way.
If it needs to be added some day, b64 encoded is very common to send binary data in a json repsonse.

balloob: The problem with sending big payloads is that your web socket connection will not be used for anything else while the big file is being send. So any event will be lost .

PR #24, opened 2021-01-11
Add integration test

balloob:
Add an integration test that spins up a server and connects to it with a client and asserts responses are correct.


Michsior14 (approved):
While it's fine for now, I would recommend to introduce `jest` and maybe `jest-websocket-mock` for testing.

balloob:
Yeah that is probably nicer. But don't underestimate my laziness 😅 This will be a great stopgap until something proper is added.

PR #27, opened 2021-01-11
Reverse CLI arg parsing

balloob:
This reverts the CLI arg parsing from #22 as it doesn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ ts-node src/bin/client.ts
 To specify different host:
 
 ```shell
-ts-node src/bin/client.ts 192.168.1.100:6000
+ts-node src/bin/client.ts ws://192.168.1.100:6000
 ```
 
 To specify that it outputs each message on a single line so it can be replayed later:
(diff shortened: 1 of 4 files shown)


robertsLando:
Wait a moment, what wasn't working?

balloob:
These were the things I noticed:

1. Doesn't fail when giving unknown keyword args.
2. Mock driver didn't work without requiring to pass positional arg.
  ```
  $ ts-node src/bin/server.ts --mock-driver
  Error: Missing path to serial port
  ```

3. Client url was split into host and port, but no longer allowed setting path. Host included protocol.

robertsLando:
> Doesn't fail when giving unknown keyword args.

This needs to be enabled in yargs parsing, I didn't enabled strict checks

> Mock driver didn't work without requiring to pass positional arg.

So the positional parameter should have been made optional or, if it's missing, check if mock-driver is present

> Client url was split into host and port, but no longer allowed setting path. Host included protocol.

This can be fixed too

PR #39, opened 2021-01-15
feat: logger and emit errors

### LINKED ISSUE #37: Allow to pass custom logger to zwavejs server options
We could add an option `logger` to ZwavejsServer `options`

### LINKED ISSUE #38: Emit listening event instead of await it
Actually we await listening from the server, we should instead emit the listening event, the same for error events to catch address in use errors

robertsLando:
Fixes #38 
Fixes #37


src/lib/server.ts
```diff
@@ -77,13 +80,13 @@ class Client {
       }
 
       throw new UnknownCommandError(msg.command);
-    } catch (err: unknown) {
+    } catch (err) {
```
Michsior14, line 83: This will implies error is `any` while we do not know the type. I would leave it as `unknown`.

src/lib/server.ts
```diff
@@ -212,32 +215,56 @@ class Clients {
 }
 interface ZwavejsServerOptions {
   port: number;
+  logger?: Logger;
+}
+
+export interface Logger {
+  error(message: string | Error, error?: Error): void;
```
Michsior14, line 222: I would say that all logger functions should define the argument lists as `...args: unknown[]`.

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
```
balloob, line 1: We should remove this file and add it to gitignore.

src/lib/server.ts
balloob: I think typing it is ok, since that shows what we will pass to it. Typescript looks at the type, not the interface that defined the type, so if someone passes a compatible logging object, it's all good. With unknown we couldn't communicate to consumers of this lib if the data we pass changes in the future.

robertsLando:
In your opinion should we remove async from start and emit the listening event instead? Like we do for error? Like a normal server does @balloob @Michsior14

balloob:
Porque no los dos?

balloob:
"Normal server" 🤷 

Async awaiting a task is absolutely preferred. It means you know the error that is raised is related to you calling that function. An error event can be emitted by any other action that is currently in progress.

src/lib/server.ts
Michsior14: While I agree on properly typing the interfaces whenever it's possible I don't see much sense to do it for logger's functions parameters, where we should be able to pass anything we want (as i.e. `console` family does it).

.gitignore
```diff
@@ -2,7 +2,6 @@ node_modules
 package-lock.json
 config.json
 /dist/
-.vscode/*
```
Michsior14, line 5: This one ignores `.vscode` folder ;)

I guess you should just remove the 4 lines bellow.

Michsior14 (approved):

.gitignore
robertsLando: @Michsior14 Yep, @balloob said to ignore the settings.json file but atm that's the only one file inside that folder so why not ignore the entire folder?

Michsior14: Yep, this is why I suggested to remove lines like `!.vscode/settings.json`. The above comment was added when you removed this line (`.vscode/*`) ;)

robertsLando: It's there... I removed it in one commit to be able to remove the folder, then in the next commit I have readded it

Michsior14: Yep I know, it's just outdated comment now.
BTW. [git rm](https://git-scm.com/docs/git-rm) is a easier way to do it ;)

robertsLando: @Michsior14 Didn't know about `git rm` command :laughing:

src/lib/server.ts
```diff
@@ -212,32 +215,64 @@ class Clients {
...
+  warn(message: string[]): void;
```
balloob, line 223: why is this one an array?

src/lib/server.ts
balloob: I honestly don't _really_ care about the logger. For all I care we keep it undefined 🤷

src/lib/server.ts
robertsLando: typo, fixing

robertsLando, line 223:
```suggestion
  warn(message: string): void;
```

balloob (approved):

PR #41, opened 2021-01-17
feat: add dockerfile

jcam:


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:erbium-buster AS build
+ENV LANG C.UTF-8
+
+# Copy data for add-on
+COPY . /usr/src/app
+WORKDIR /usr/src/app
+RUN yarn install
```
Michsior14, line 7:
```suggestion
RUN npm install
RUN npm run build
```

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+FROM node:erbium-buster-slim
+COPY --from=build /usr/src/app /usr/src/app
+WORKDIR /usr/src/app
+EXPOSE 3000
+USER root
+CMD [ "node_modules/.bin/ts-node", "cli/server.ts", "/dev/ttyUSB0" ]
```
Michsior14, line 14:
```suggestion
CMD [ "node", "dist/bin/server.js", "/dev/ttyUSB0" ]
```

Michsior14 (changes requested):

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 package-lock.json
 config.json
+yarn.lock
```
Michsior14, line 4: We don't need yarn.lock here anymore :)

.dockerignore
```diff
@@ -0,0 +1,12 @@
...
+*.md
+docs
+Dockerfile
+.vscode
+.nyc_output
+.coverage
+pkg
+test
+node_modules
+yarn.lock
```
Michsior14, line 12: You could add `package-lock.json` as well.

Michsior14 (approved):

jcam:
Ah, I was modifying an old version and somehow had my git origins set up wrong so I didn't realize it. Rebasing...

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN npm install
+RUN npm run build
+RUN npm prune --production
+
+FROM node:erbium-buster-slim
+COPY --from=build /usr/src/app /usr/src/app
+WORKDIR /usr/src/app
+EXPOSE 3000
+USER root
+CMD [ "node", "dist/bin/server.js", "/dev/ttyUSB0" ]
```
balloob, line 16: This makes no sense to hardcode to this path. People need to get their serial stick into Docker.

balloob:
What would be the use of adding this Dockerfile ?

jcam:
It is the dockerfile that I use to generate a package for deploying to my nuc. If not useful, close it out. I was just offering it in case others would find it useful, rather than figuring out how to build it themselves. Could also put it in a docs/examples folder or something

Michsior14:
> It is the dockerfile that I use to generate a package for deploying to my nuc. If not useful, close it out. I was just offering it in case others would find it useful, rather than figuring out how to build it themselves. Could also put it in a docs/examples folder or something

I could also see that happening on my end. I tent to keep zwave controller separately of HassOS.

Dockerfile
balloob: And also please add a section to the README how to use this Docker file.

jcam:
Closing for now in favor of https://github.com/kpine/dockerfiles/tree/main/zwave-js-server

PR #42, opened 2021-01-17
fix: add lock file and mention production usage

Michsior14:


MartinHjelmare (approved):

README.md
```diff
@@ -4,6 +4,8 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
 
 ## Trying it out
 
+Keep in mind that on production you want to execute `npm run build` and use `node` instead of `ts-node` for all bellow commands.
```
balloob, line 7:
```suggestion
Keep in mind that on production you want to execute `npm run build` and use `node` instead of `ts-node` for all below commands.
```

.gitignore
```diff
@@ -1,5 +1,4 @@
 node_modules
-package-lock.json
```
balloob, line 2: So this project can both be used as a lib and as a standalone server. When installed as a lib (ie as part of ZJS2MQTT), the package-lock file will be ignored.

README.md
balloob: I don't htink that this sentence makes sense.

In production, when installed from `npm`, it is already JS and they should use the CLIs that we install: 

https://github.com/zwave-js/zwave-js-server/blob/master/package.json#L6-L9

Michsior14: That's true if installed via `npm`, but it might be worth to mention that `ts-node` is not preferred for running it in long term if someone decide to clone.

.gitignore
Michsior14: If we ignore `package-lock.json` from git we basically allow to unintentionally bump the dependencies during `npm i` on github actions and contributors machines. This can lead to failed tests/builds. 
In the end this file **won't** be published to [npm by default](https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json), so it won't be a problem for the lib consumers.

balloob: We should publish it to npm, for when it's installed as a CLI. 

When an npm package is installed as a dependency, the package-lock file is already ignored anyway.

Michsior14: You can publish `npm-shrinkwrap.json` only. I will add the proper command to the github action.

kpine: Sorry to jump in, but when I install this from npm (`npm install @zwave-js/server`), it's currently pulling in zwave-js 6.0.0-beta.3 as a dependency. Is that what this conversation is about? I don't think I should have that version as a dependency yet...

README.md
```diff
@@ -4,6 +4,8 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
 
 ## Trying it out
 
+Keep in mind that on production you want to execute `npm run build` and use `node` instead of `ts-node` for all below commands.
```
robertsLando, line 7: That's not correct. If you install this using the command `npm install -g @zwave-js/server` you will be able to run this by simply call `zwave-server /dev/ttyACM0` and `zwave-client` taken from: https://github.com/zwave-js/zwave-js-server/blob/master/package.json#L7

robertsLando (changes requested):

README.md
balloob, line 7:
```suggestion
These instructions are for development only. These CLIs will be available as `zwave-server` and `zwave-client` after installing the NPM package.
```

balloob (approved):

PR #43, opened 2021-01-17
Bump zwave-js to 6.0 RC

jcam:
Upstream has a release candidate that's ready for further testing


balloob:
We only upgrade to [officially tagged releases](https://github.com/zwave-js/node-zwave-js/releases).

PR #61, opened 2021-01-24
server.ts - fix replace to use global regex

itewk:
replace only replaces the first instance, update to replace all instances using regex


marcelveldt:
Oops, good catch!

marcelveldt (approved):

PR #62, opened 2021-01-24
Include metadata and ccVersion in added and notification events

marcelveldt:
Discussed here:
https://github.com/home-assistant-libs/zwave-js-server-python/issues/50

Value's metadata is only sent in the full state (at initial connection) but never when new values are added (e.g. a node completes interview after the server init or a new node was added). That is problematic as we need the metadata in the python library (and HA integration).

We can alter the python lib to ask for the metadata through the getValueMetaData command call but this feels better so it is more in line with the initial state object and second reason is that we have to refactor the python lib as commands are async (and the event handlers are not now).

The state (and value added event) now also includes the ccVersion as this is important metadata to know, for example to solve the targetValue issue in HA light platform.

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -136,7 +136,8 @@ export class EventForwarder {
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
           // include metadata in the response
-          args.metadata = node.getValueMetadata(args);
+          const metadata = node.getValueMetadata(args);
+          args.metadata = metadata;
           notifyNode(changedNode, event, { args });
         });
       }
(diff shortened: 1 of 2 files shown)


src/lib/forward.ts
```diff
@@ -133,6 +131,21 @@ export class EventForwarder {
       }
     }
 
+    {
+      const events: ZWaveNodeEvents[] = ["value added", "value notification"];
```
balloob, line 135: If we already add the info to `value added`, we don't need to include it in `value notification` right ?

marcelveldt: Value notification is a special stateless one. It will never be added ;-)

balloob: TIL

balloob (approved):

PR #63, opened 2021-01-24
Development config

marcelveldt:
a few changes for development with VS Code on MacOS


marcelveldt:
This ESLint config is more strict. What to do ? Adjust code so the linting works reliable or loosen it ?

.eslintrc.js
```diff
@@ -10,4 +10,7 @@ module.exports = {
     sourceType: "module",
   },
   plugins: ["@typescript-eslint"],
+  rules: {
+    "no-unused-vars": "off",
```
robertsLando, line 14: @marcelveldt My suggestion is to keep this on and add the ignore lines where needed. This rule helps

robertsLando (changes requested):

marcelveldt:
We should address all warnings that the linter has found with these stricter linting config.
I'll do that when I get bored next couple of days ;-)

src/lib/error.ts
```diff
@@ -1,7 +1,8 @@
+/* eslint-disable no-unused-vars */
```
robertsLando, line 1: Based on https://stackoverflow.com/questions/57802057/eslint-configuring-no-unused-vars-for-typescript

YOu could remove all this lines, just try to update `typescript-eslint/recommended` to latest version

robertsLando (changes requested):

src/lib/error.ts
marcelveldt: Yeah, I just figured out this is fixed in the latest version

marcelveldt: @robertsLando one more question. It's now forcing a tabwidth of 2 spaces, is that correct?

marcelveldt:
Why is CI still testing with the old config? All tests run fine on my machine. Looks as if the CI task is using some cached info or is not using the linter config from this PR.

marcelveldt:
This one is finally done. It's using the recommended setup of ESlint with the prettier plugin and typescript support now.

src/lib/error.ts
robertsLando: Dunno if it's correct or not I mean it's a matter of personal choice, you can choose the preferred indentation

marcelveldt: Yeah, I've set it back to 2. 
Kind of used to 4 in the python world ;-)

.eslintrc.js
```diff
@@ -1,13 +1,20 @@
...
-  parserOptions: {
-    ecmaVersion: 11,
-    sourceType: "module",
+    node: true
   },
-  plugins: ["@typescript-eslint"],
+  extends: [
+    "plugin:@typescript-eslint/recommended", // Uses the recommended rules from the @typescript-eslint/eslint-plugin
+    "prettier/@typescript-eslint", // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
+    "plugin:prettier/recommended" // Enables eslint-plugin-prettier and eslint-config-prettier. This will display prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
```
balloob, line 14: I am against this addition. Prettier shouldn't show up as an error. We already automatically fix it before commit and people can set up format on save.

balloob: It pollutes the editor when writing.

marcelveldt: This change prefers ESLint with prettier plugin so it should be better for use with VS Code as it's the recommended way according to the VS Code team. What do you suggest ?

.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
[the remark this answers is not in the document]
marcelveldt: I can leave out the prettier config so it will use the defaults ;-)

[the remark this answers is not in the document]
marcelveldt: Having this file in place will make sure that no VS-Code user configured file is messing with the defaults though.

.eslintrc.js
balloob: It's super annoying to see ESLine errors and warnings that you know are going to be automatically formatted away when you hit save. I vote against it .

.prettierrc.js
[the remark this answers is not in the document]
balloob: We have a CI check to enforce that ;) I would prefer we just go with defaults.

.eslintrc.js
marcelveldt: Ah, now I get what you mean now. My only reason for this PR was to enforce default formatting rules but I don't care how :-)

PR #67, opened 2021-01-26
Temporary fix for getEndpoint exception

marcelveldt:
@MartinHjelmare this fixes the crashing at startup but I don't like it :-)
We can merge this in for now but I'm going to submit another PR to do all the state dumping in smaller pieces (node per node)

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -138,9 +138,17 @@ export class EventForwarder {
           // include metadata and ccVersion in the response
           const metadata = node.getValueMetadata(args);
           args.metadata = metadata;
-          args.ccVersion =
-            node.getEndpoint(args.endpoint).getCCVersion(args.commandClass) ||
-            node.getEndpoint(0).getCCVersion(args.commandClass);
+          try {
+            args.ccVersion =
+              node.getEndpoint(args.endpoint).getCCVersion(args.commandClass) ||
+              node.getEndpoint(0).getCCVersion(args.commandClass);
+          } catch {
+            // TEMPORARY FIX: race condition - node is being interviewed and endpoint is not yet available
+            // https://github.com/zwave-js/node-zwave-js/issues/1499
+            args.ccVersion = node
+              .getEndpoint(0)
+              .getCCVersion(args.commandClass);
+          }
           notifyNode(changedNode, event, { args });
         });
       }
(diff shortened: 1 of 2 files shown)


src/lib/forward.ts
```diff
@@ -138,9 +138,17 @@ export class EventForwarder {
           // include metadata and ccVersion in the response
           const metadata = node.getValueMetadata(args);
           args.metadata = metadata;
-          args.ccVersion =
-            node.getEndpoint(args.endpoint).getCCVersion(args.commandClass) ||
-            node.getEndpoint(0).getCCVersion(args.commandClass);
+          try {
+            args.ccVersion =
+              node.getEndpoint(args.endpoint).getCCVersion(args.commandClass) ||
+              node.getEndpoint(0).getCCVersion(args.commandClass);
```
robertsLando, line 144:
```suggestion
              node.getEndpoint(args.endpoint)?.getCCVersion(args.commandClass) ||
              node.getEndpoint(0)?.getCCVersion(args.commandClass);
```

In this way you can remove the try/catch

robertsLando (changes requested):

src/lib/forward.ts
marcelveldt: isn't endpoint 0 always available as it's the node itself ?

robertsLando: > isn't endpoint 0 always available as it's the node itself ?

Maybe yes, don't remember, but it doesn't add any overhead so I would keep the `?` :)

marcelveldt:
@robertsLando ready for a final review before merge ?

robertsLando (approved):
LGTM

robertsLando:
@marcelveldt BTW I would add those infos only when node is ready, but as a temporary fix this is ok :)

marcelveldt:
I've just discussed this on discord with @balloob and we're about to change the logic a bit so we well indeed only request all values info when the node is ready. That will solve this headache issues ;-)

robertsLando:
> That will solve this headache issues ;-)

Good! FYI there is also this: https://github.com/zwave-js/zwavejs2mqtt/issues/347

Actually it's a bug, in poor words it could happen that the manufacturer CC isn't available on node ready event.

marcelveldt:
temporary fix not needed as I'm about to publish a new PR with the definitive fix for this issue.

PR #68, opened 2021-01-26
Only dump full node values if the node is ready

marcelveldt:
This is the real fix for issue discussed today where value events for a node are fired before the node is ready.

- The dumpState will only dump all values for a node if the node is ready. If the node is not ready, the values will be an empty array.
- Value events (value added, value changed) will not be forwarded if the node is not is a ready state.
- When the node hits ready, a full nodeState will be dumped into the event, including all value infos.

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -108,15 +108,11 @@ export class EventForwarder {
         ...extra,
       });
 
-    {
-      const events: ZWaveNodeEvents[] = ["interview completed", "ready"];
-      for (const event of events) {
-        node.on(event, (changedNode: ZWaveNode) => {
-          const nodeState = dumpNode(changedNode);
-          notifyNode(changedNode, event, { nodeState });
-        });
-      }
-    }
+    node.on("ready", (changedNode: ZWaveNode) => {
+      // Dump full node state on ready event
+      const nodeState = dumpNode(changedNode);
+      notifyNode(changedNode, "notification", { nodeState });
+    });
 
     {
       const events: ZWaveNodeEvents[] = ["wake up", "sleep", "dead", "alive"];
@@ -128,7 +124,10 @@ export class EventForwarder {
     }
 
     {
-      const events: ZWaveNodeEvents[] = ["value removed", "interview failed"];
+      const events: ZWaveNodeEvents[] = [
+        "interview completed",
+        "interview failed",
+      ];
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
           notifyNode(changedNode, event, { args });
@@ -141,6 +140,7 @@ export class EventForwarder {
         "value updated",
         "value added",
         "value notification",
+        "value removed",
       ];
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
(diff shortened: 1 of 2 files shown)


src/lib/forward.ts
```diff
@@ -119,33 +128,33 @@ export class EventForwarder {
...
-            node.getEndpoint(0).getCCVersion(args.commandClass);
+          // only forward value events for ready nodes
+          if (!changedNode.ready) return;
           notifyNode(changedNode, event, { args });
         });
       }
     }
 
+    node.on(
+      "value notification",
```
robertsLando, line 151: @marcelveldt  Why don't add this to the `ZwaveNodeEvents` ? I think that you should check for the node to be ready here too

marcelveldt: I did that because we dump that to a full state too but I guess this doesn't add any noticeable overhead. Adjusted it.

robertsLando (approved):
LGTM

src/lib/forward.ts
```diff
@@ -103,9 +111,10 @@ export class EventForwarder {
     {
       const events: ZWaveNodeEvents[] = ["interview completed", "ready"];
       for (const event of events) {
-        node.on(event, (changedNode: ZWaveNode) =>
-          notifyNode(changedNode, event)
-        );
+        node.on(event, (changedNode: ZWaveNode) => {
+          const nodeState = dumpNode(changedNode);
```
balloob, line 115: The description mention doing this only for `ready` but here you're doing the same for `interview completed`. Is that expected?

marcelveldt: Well, I did that just in case to be complete. So for both ready and interview completed the full node is dumped.
"interview completed" fires after "ready"

robertsLando: `interview completed` means that all value metadata are updated. So it makes sense

robertsLando: Just keep in mind that, for battery devices, the ready event could take so long before its emitted, even more for the interview completed

marcelveldt: We might even want to process "interview completed" in the lib to do our magic instead of ready

marcelveldt: > Just keep in mind that, for battery devices, the ready event could take so long before its emitted, even more for the interview completed

Yeah, that's why we send the basic node info before that so we can at least show the node info that is already present and create HA entities after all metadata is there.

robertsLando: @AlCalzone could make better suggestions then me here

marcelveldt: I've been watching the events just now and "ready" is emitted before "interview completed" and some values get added in between the two. So I guess we should consider "interview completed" as our "ready to do something" rather than "node ready" ?

AlCalzone: Depends on the use case. After the very first interview, `ready` and `interview completed` are emitted back to back
During the following interviews, `ready` is emitted very soon, since we have all information about the node to interact with it and all metadata for all values. `interview completed` is then raised when all values themselves are updated, but that might take very long.
I advise listening for `ready` and treating everything after that as updated values.

marcelveldt: Thanks! So when "ready" is emitted all metadata etc. will be there. No need to wait for "interview completed" ?

AlCalzone: nope, only if you absolutely need up to date values

marcelveldt: Agreed. This is already handled by the fact that value events are emitted after the node is ready.
All clear now then.

marcelveldt:
@balloob ready for merge ?

PR #70, opened 2021-01-26
Add poll value command

marcelveldt:
This command was added in Zwave JS 6.1.0

Might come in handy for HA so users can eventually poll values manually if that is really required.
I'll expose it with a service call (with a big fat warning)

### DIFF
--- a/src/lib/node/message_handler.ts
+++ b/src/lib/node/message_handler.ts
@@ -32,8 +32,8 @@ export class NodeMessageHandler {
         await node.abortFirmwareUpdate();
         return {};
       case NodeCommand.pollValue:
-        const result = await node.pollValue<any>(message.valueId);
-        return { result };
+        const value = await node.pollValue<any>(message.valueId);
+        return { value };
       default:
         throw new UnknownCommandError(command);
     }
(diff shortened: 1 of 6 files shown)


src/lib/node/message_handler.ts
```diff
@@ -31,6 +31,8 @@ export class NodeMessageHandler {
       case NodeCommand.abortFirmwareUpdate:
         await node.abortFirmwareUpdate();
         return {};
+      case NodeCommand.pollValue:
+        return await node.pollValue(message.valueId);
```
balloob, line 35: The return value should always be wrapped in an object. `{ value: X }`. 

Also make sure that the result value can be JSON serialized.

marcelveldt: I was not sure if this actually returns the value as the docs are a bit unclear. I'll inspect the code to find out.

marcelveldt:
I've adjusted it. As the value's type is unknown, I've set it to any.

src/lib/node/message_handler.ts
```diff
@@ -31,6 +31,9 @@ export class NodeMessageHandler {
       case NodeCommand.abortFirmwareUpdate:
         await node.abortFirmwareUpdate();
         return {};
+      case NodeCommand.pollValue:
+        const result = await node.pollValue<any>(message.valueId);
```
balloob, line 35: So it looks like this returns the updated value. So this should be called `value` and not `result` ?

marcelveldt: It *may* return the value if the node responds and polling is supported in the CC.
But yeah, I could rename it to value

balloob (approved):

PR #72, opened 2021-01-29
Small fix for value (notification) events

marcelveldt:
Value notification value is already provided in the event.
This change will make sure that the message signature is always consistent for all value events, where the value attribute always contains the most recent value.

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -124,8 +124,8 @@ export class EventForwarder {
       const events: ZWaveNodeEvents[] = [
         "value updated",
         "value added",
-        "value removed",
         "value notification",
+        "value removed",
       ];
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -121,16 +121,24 @@ export class EventForwarder {
     }
 
     {
-      const events: ZWaveNodeEvents[] = [
-        "value updated",
-        "value added",
-        "value notification",
-        "value removed",
-      ];
+      const events: ZWaveNodeEvents[] = ["value updated", "value removed"];
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
           // only forward value events for ready nodes
           if (!changedNode.ready) return;
+          // vakue updated/removed events are forwarded as-is
+          notifyNode(changedNode, event, { args });
+        });
+      }
+    }
+
+    {
+      const events: ZWaveNodeEvents[] = ["value added", "value notification"];
+      for (const event of events) {
+        node.on(event, (changedNode: ZWaveNode, args: any) => {
+          // only forward value events for ready nodes
+          if (!changedNode.ready) return;
+          // value added/notification should contain metadata, use dumpValue
           const valueState = dumpValue(changedNode, args);
           notifyNode(changedNode, event, { args: valueState });
         });
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -4,10 +4,9 @@ import {
   NodeStatus,
   ZWaveNode,
   ZWaveNodeEvents,
-  ZWaveNodeValueNotificationArgs,
 } from "zwave-js";
 import { OutgoingEvent } from "./outgoing_message";
-import { dumpNode, dumpValue, dumpValueNotification } from "./state";
+import { dumpNode, dumpValue } from "./state";
 
 export class EventForwarder {
   /**
@@ -126,6 +125,7 @@ export class EventForwarder {
         "value updated",
         "value added",
         "value removed",
+        "value notification",
       ];
       for (const event of events) {
         node.on(event, (changedNode: ZWaveNode, args: any) => {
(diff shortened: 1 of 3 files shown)


src/lib/state.ts
```diff
@@ -50,6 +51,16 @@ export const dumpValue = (
   return valueState;
 };
 
+export const dumpValueNotification = (
```
balloob, line 54: If this is only used in the forwarder and it depends on the `valueArgs`, there is no need to extract it in a function.

src/lib/state.ts
```diff
@@ -50,6 +51,16 @@ export const dumpValue = (
...
+  node: ZWaveNode,
+  valueArgs: ZWaveNodeValueNotificationArgs
+): ValueState => {
+  const valueState = valueArgs as ValueState;
+  valueState.metadata = node.getValueMetadata(valueArgs);
+  valueState.ccVersion = 0; // no need for ccVersion in these stateless events (for now)
```
balloob, line 60: If there is no need, let's not include it?

marcelveldt:
I've simplified it a bit as it makes sense that the ValueState object is more generic.
The value key always contains the right value now. This is also easier in the python lib.
Strictly taken this should done with if-else (to account for value 0) but this looks cleaner imo and the getValue call is light.

src/lib/state.ts
marcelveldt: I wanted the object to have the same signature as ValueState. I've now changed it so it still uses the same dumpValue function.

src/lib/state.ts
marcelveldt: Decided to keep it in for consistency reasons.

src/lib/state.ts
```diff
@@ -36,17 +36,23 @@ function getNodeValues(node: ZWaveNode): ValueState[] {
...
 
-export const dumpValue = (
-  node: ZWaveNode,
-  valueId: TranslatedValueID
-): ValueState => {
-  const valueState = valueId as ValueState;
-  valueState.metadata = node.getValueMetadata(valueId);
-  valueState.value = node.getValue(valueId);
+export const dumpValue = (node: ZWaveNode, valueArgs: any): ValueState => {
+  const valueState = valueArgs as ValueState;
```
balloob, line 40: why is this necessary and can't we put that in the function signature?

src/lib/forward.ts
```diff
@@ -124,8 +124,8 @@ export class EventForwarder {
       const events: ZWaveNodeEvents[] = [
         "value updated",
         "value added",
-        "value notification",
         "value removed",
+        "value notification",
```
balloob, line 128: This is now not a change?

src/lib/state.ts
marcelveldt: I want to make sure that all value related messages generate the same valueState message.
So, the value can come from an event but also from enumerating the valueID's on the node.
It's either duplicating code or have one place that contains this logic.

src/lib/state.ts
```diff
@@ -36,17 +36,23 @@ function getNodeValues(node: ZWaveNode): ValueState[] {
...
 }
 
-export const dumpValue = (
-  node: ZWaveNode,
-  valueId: TranslatedValueID
-): ValueState => {
-  const valueState = valueId as ValueState;
-  valueState.metadata = node.getValueMetadata(valueId);
-  valueState.value = node.getValue(valueId);
+export const dumpValue = (node: ZWaveNode, valueArgs: any): ValueState => {
```
balloob, line 39: Did you test this? It seems like we're changing the signature and miss out on typing ?

src/lib/forward.ts
marcelveldt: No, with the consolidating the dumpValue this can all be handled

src/lib/state.ts
marcelveldt: Yes, I did test this. valueArgs can be either ValueChangedArgs, ValueAddedArgs etc. etc so it was either a lot of OR's types or abusing any.

src/lib/state.ts
marcelveldt: FYI: we add the ccVersion and metadata to all value messages.

marcelveldt: But if you want I could also just do the specifics in each function and use the dumpValue only for the dumpState

src/lib/state.ts
marcelveldt: adjusted

src/lib/state.ts
balloob: I'm just scared that we accidentally break something and typing won't find it. Is this line still necessary even with the translated value id type?

marcelveldt: Yes because we “convert” the incoming ValueID to ValueState, which includes value, metadata and ccVersion.

this line was already here btw, before this change.

src/lib/state.ts
```diff
@@ -38,15 +43,33 @@ function getNodeValues(node: ZWaveNode): ValueState[] {
...
+      node
+        .getEndpoint(valueArgs.endpoint)
+        ?.getCCVersion(valueArgs.commandClass) ||
+      node.getEndpoint(0).getCCVersion(valueArgs.commandClass),
+  };
+
+  // make sure that value attribute always holds correct value
+  if ("newValue" in valueArgs) {
+    valueState.value = valueArgs.newValue;
+  } else if ("value" in valueArgs) {
```
marcelveldt, line 69: small typo, should be !"value" in valueArgs. Only if there is value in the valueArgs we retrieve it with node.getValue

marcelveldt: adjusted

marcelveldt:
tested this branch on my test mesh. Works great.
Waiting on final approval from @balloob before merge.

balloob (approved):

marcelveldt:
Changes made like discussed on discord.
The events "value_updated" and "value_removed" will be forwarded as-is. "value_added" and "value_notification" will be extended with metadata and ccVersion, same for a value that is dumped in the dumpNode function.

balloob (approved):

PR #81, opened 2021-02-02
Fix value events

marcelveldt:
As discussed with @AlCalzone we missed the metadata updated event and so we do not need the workaround.
This change will:

- Forward all value events as-is
- Only in the dumpState/dumpNode function a full value object, including metadata and ccVersion will be sent
- metadata updated event will be forwarded, which can be used to update metadata of values
- This will/should fix the issue that the value's value is not present in Scene Activation events.


robertsLando (approved):

AlCalzone (approved):

balloob (approved):
This is a hot code path for PRs 😅

PR #103, opened 2021-02-09
Move zwave-js to peer/dev deps

### LINKED ISSUE #99: Inconsistency in zwave-js dependencies
I'm not sure if this is the fault of the server module or zwavejs2mqtt, but here it goes...
We have two copies of the zwave-js dependencies in the Docker container.

https://github.com/zwave-js/node-zwave-js/issues/1615#issuecomment-775209809
```
/usr/src/app # find . | grep tz69
./node_modules/@zwave-js/server/node_modules/@zwave-js/config/config/devices/0x0118/tz69.json
./node_modules/@zwave-js/config/config/devices/0x0118/tz69.json
```

This is going to cause some headaches. /cc @robertsLando @balloob @marcelveldt

balloob:
This moves zwave-js to peer/dev deps. 

**Breaking change:** If you are deploying containers that just install this package, you now need to install the npm package `zwave-js` too.

Fixes #99 

CC @AlCalzone @robertsLando


robertsLando (approved):
No changes need on my side

PR #116, opened 2021-02-15
Add driver command to update log config

### LINKED ISSUE #115: Add command to update logging configuration
This was merged and released a litttle while ago: https://github.com/zwave-js/node-zwave-js/pull/1754

It would be great to add support for this in the server as well as in the Python library so we can add a service in Home Assistant to update the log level. This would eliminate one of the common reasons why users currently restart server.

raman325:
Related to https://github.com/zwave-js/zwave-js-server/issues/115 and https://github.com/zwave-js/node-zwave-js/pull/1754. This will allow us to add support to update the log configuration in the Python library and then expose a service in Home Assistant to update the log levels without having to restart the server.

The version bump is a breaking change to the add-on - updating the add-on would depend on https://github.com/home-assistant/core/pull/46355 in order for `zwave_js` thermostats in Home Assistant to continue to function.

Resolves #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,16 @@ interface {
 
 ## Client commands
 
+### Start Listening for events
+
 ```ts
 interface {
   messageId: string;
   command: "start_listening";
 }
 ```
+
+### Set value on a node
 
 ```ts
 interface {
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ interface {
 
 ## Client commands
 
-### Start Listening for events
+### Start listening to events
 
 ```ts
 interface {
@@ -127,6 +127,10 @@ interface {
   value: any
 }
 ```
+
+### Update the logging configuration
+
+> NOTE: You must provide at least one key/value pair as part of `config`
 
 ```ts
 interface {
--- a/README.md
+++ b/README.md
@@ -124,6 +124,21 @@ interface {
 }
 ```
 
+```ts
+interface {
+  messageId: string;
+  command: "update_log_config";
+  config: {
+    enabled?: boolean;
+    level?: number;
+    transports?: Transport[];
+    logToFile?: boolean;
+    filename?: string;
+    forceConsole?: boolean;
+  }
+}
+```
+
 ## Authentication
 
 Z-Wave JS Server does not handle authentication and allows all connections to the websocket API. If you want to add authentication, add authentication middleware to your Express instance or run NGINX in front of Express instance.
--- a/README.md
+++ b/README.md
@@ -139,7 +139,6 @@ interface {
   config: {
     enabled?: boolean;
     level?: number;
-    transports?: Transport[];
     logToFile?: boolean;
     filename?: string;
     forceConsole?: boolean;
--- a/src/lib/incoming_message.ts
+++ b/src/lib/incoming_message.ts
@@ -1,3 +1,5 @@
+import { LogConfig } from "@zwave-js/core";
+import { DeepPartial } from "@zwave-js/shared";
 import { IncomingMessageController } from "./controller/incoming_message";
 import { IncomingCommandBase } from "./incoming_message_base";
 import { IncomingMessageNode } from "./node/incoming_message";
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "ws": "^7.4.2"
   },
   "peerDependencies": {
-    "zwave-js": "^6.2"
+    "zwave-js": "^6.3"
```
MartinHjelmare, line 28: Let's do separate PRs for bumping zwave-js, the log feature addition and bumping the server version.

MartinHjelmare (changes requested):

raman325:
I will rebase once #118 is merged

MartinHjelmare:
#118 is merged. :+1:

MartinHjelmare:
I'd like some TS/JS eyes on this PR.

raman325:
> I'd like some TS/JS eyes on this PR.

Do we want to bump the server version now, or wait until this is merged?

MartinHjelmare:
We can do a release now of the server and have this PR go in a later release.

README.md
```diff
@@ -124,6 +128,25 @@ interface {
...
+> NOTE: You must provide at least one key/value pair as part of `config`
+
+```ts
+interface {
+  messageId: string;
+  command: "update_log_config";
+  config: {
+    enabled?: boolean;
+    level?: number;
+    transports?: Transport[];
```
balloob, line 142: Can we expand the `Transport` type, not sure what it is here.

src/lib/outgoing_message.ts
```diff
@@ -29,6 +29,8 @@ interface OutgoingResultMessageError {
 
 export type ResultTypes = {
   start_listening: { state: ZwaveState };
+} & {
+  update_log_config: Record<string, never>;
 } & NodeResultTypes &
```
balloob, line 34: I think that this should be an `|` ?

```suggestion
} | {
  update_log_config: Record<string, never>;
} | NodeResultTypes |
```

I also think that we want to extract the 2 anonymous defined interfaces into named interfaces.

src/lib/incoming_message.ts
```diff
@@ -7,7 +9,14 @@ interface IncomingCommandStartListening extends IncomingCommandBase {
   command: "start_listening";
 }
 
+interface IncomingCommandUpdateLogConfig extends IncomingCommandBase {
+  messageId: string;
+  command: "update_log_config";
+  config: DeepPartial<LogConfig>;
```
balloob, line 15: Why is this `DeepPartial` and not `Partial` ? Can partial transport configs be defined ?

raman325: I used the type that is expected upstream: https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/driver/Driver.ts#L590

Will `Partial` work instead?

I don't know anything about the `Transport` part of the config but it seems that is mainly for internal use, e.g. it shouldn't be updated externally (https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/log/shared.ts#L129)

src/lib/outgoing_message.ts
raman325: I followed the existing convention by using `&` but `|` makes more sense if I understand this correctly. How was this not broken before?

README.md
raman325: I'm wondering if we should just drop it altogether (drop it from the `node-zwave-js` response before serialization)? As I mentioned in another PR, I think this is managed entirely internally. From what I can tell, transports are really just a stream writer that is created based on the other parameter values.

balloob: yeah, in that case let's not expose more than we need right now for HA.

src/lib/outgoing_message.ts
balloob: I think it's just used for typing but probably not very useful haha

raman325: looks like `&` is right, `|` results in type errors

src/lib/incoming_message.ts
raman325: I looked into `DeepPartial` and `Partial` and since we have decided not to support updating transports, I don't think `DeepPartial` is needed

raman325:
Since we are on the latest version of `node-zwave-js`, this PR can be merged with the other one. Closing this one since that one is more complete

PR #117, opened 2021-02-15
Add commands to get and update log config

### LINKED ISSUE #115: Add command to update logging configuration

raman325:
Related to #115 and zwave-js/node-zwave-js#1754. This will allow us to add support to get and update the log configuration in the Python library and then expose a UI in Home Assistant to get and update the log levels without having to restart the server.

Resolves #115

### DIFF
--- a/src/lib/outgoing_message.ts
+++ b/src/lib/outgoing_message.ts
@@ -30,7 +30,7 @@ interface OutgoingResultMessageError {
 export type ResultTypes = {
   start_listening: { state: ZwaveState };
 } & {
-  update_log_config: { success: boolean };
+  update_log_config: Record<string, never>;
 } & NodeResultTypes &
   ControllerResultTypes;
 
--- a/src/lib/outgoing_message.ts
+++ b/src/lib/outgoing_message.ts
@@ -29,6 +29,8 @@ interface OutgoingResultMessageError {
 
 export type ResultTypes = {
   start_listening: { state: ZwaveState };
+} & {
+  update_log_config: { success: boolean };
 } & NodeResultTypes &
   ControllerResultTypes;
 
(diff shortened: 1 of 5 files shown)


src/lib/outgoing_message.ts
```diff
@@ -31,6 +32,8 @@ export type ResultTypes = {
   start_listening: { state: ZwaveState };
 } & {
   update_log_config: Record<string, never>;
+} & {
```
balloob, line 35: Extract this too .

src/lib/server.ts
```diff
@@ -78,6 +78,13 @@ class Client {
         return;
       }
 
+      if (msg.command === "get_log_config") {
+        this.sendResultSuccess(msg.messageId, {
+          config: this.driver.getLogConfig(),
```
balloob, line 83: Is everything returned from this function JSON serializable or should certain parts be manually converted (ie `Transport`)?

raman325: See my comment here: https://github.com/zwave-js/zwave-js-server/pull/116#discussion_r577374393

marcelveldt:
Nothing is blocking this PR anymore right ?

raman325:
Nope, forgot to switch it out of draft 😅

marcelveldt:
this requires a new version of zwave-js right ? Should we bump the requirement ?

marcelveldt (approved):
Looks good!

raman325:
Yes we will need a bump, the convention has been to bump the version in a separate PR

marcelveldt:
Yes, so do the bump PR right after merging this because this one depends on it.

PR #128, opened 2021-02-21
Add node command to send raw Config CC values

raman325:
One of the goals for the HA `set_config_parameter` service is to allow users to send raw values for bitwise parameters that are either split out into partial parameters if they are defined in the device config or as a single value without any metadata if they are not. This PR adds a node command that we can use for those use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,10 @@ interface {
 }
 ```
 
+### Controller level commands
+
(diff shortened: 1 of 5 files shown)


balloob (approved):
looks good. Ok to merge when README updated.

raman325:
@balloob seemed like there was an opportunity to improve the docs along with adding this command, so I went ahead and did so - let me know if you would like it to be formatted differently

marcelveldt (approved):
looks fine to me!

PR #136, opened 2021-02-23
Extend dump to include full deviceclass object

### LINKED ISSUE #75: Add supported CommandClasses to dumpNode
It would be nice to also list the supported commandclasses of a node in the dump function.
Usecase for this is to know upfront if a device supports e.g. Central Scene CC to create device triggers in the HA integration.

### LINKED ISSUE #76: Extend deviceclass dump
The dumpNode function currently only dumps the deviceClass keys which are the string representations.
We should extend this so that both the deviceclass integer value and the string is exposed.

marcelveldt:
**BREAKING CHANGE**

DeviceClass strings tend to change sometimes which is annoying as our discovery scheme in HA depends a lot on these deviceclass id's.

This change replaces the deviceClass to a full object containing both the key and label.

Discussed with @MartinHjelmare that we should do this breaking change before HA beta release (and adjust both integration and the lib for this change).

Closes #76
Closes #75 

The dump will look like this after this change:

```
"deviceClass": {
            "basic": {
              "key": 2,
              "label": "Static Controller"
            },
            "generic": {
              "key": 2,
              "label": "Static Controller"
            },
            "specific": {
              "key": 1,
              "label": "PC Controller"
            },
            "mandatorySupportedCCs": [
              
            ],
            "mandatoryControlledCCs": [
              32
            ]
          },
```

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -5,14 +5,38 @@ import {
   Endpoint,
   TranslatedValueID,
   ValueMetadata,
+  DeviceClass,
 } from "zwave-js";
+
+type Modify<T, R> = Omit<T, keyof R> & R;
 
 export interface ZwaveState {
   controller: Partial<ZWaveController>;
-  nodes: Partial<ZWaveNode>[];
+  nodes: NodeState[];
 }
 
 interface EndpointState extends Partial<Endpoint> {}
+
+interface DeviceClassState
+  extends Partial<
+    Modify<
+      DeviceClass,
+      {
+        basic: {
+          key: number;
+          label: string;
+        };
+        generic: {
+          key: number;
+          label: string;
+        };
+        specific: {
+          key: number;
+          label: string;
+        };
+      }
+    >
+  > {}
 
 interface ValueState extends TranslatedValueID {
   metadata: ValueMetadata;
@@ -20,10 +44,17 @@ interface ValueState extends TranslatedValueID {
   value?: any;
 }
 
-interface NodeState extends Partial<ZWaveNode> {
-  endpoints: EndpointState[];
-  values: ValueState[];
-}
+interface NodeState
+  extends Partial<
+    Modify<
+      ZWaveNode,
+      {
+        deviceClass: DeviceClassState;
+        endpoints: EndpointState[];
+        values: ValueState[];
+      }
+    >
+  > {}
 
 function getNodeValues(node: ZWaveNode): ValueState[] {
   if (!node.ready) {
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
What's the reason for not including the full state as described here?
https://zwave-js.github.io/node-zwave-js/#/api/node?id=deviceclass

marcelveldt:
Well because we will have to replicate that entire object and no use ?

marcelveldt:
Like just discussed on discord with @AlCalzone the json serializations of most objects only return the string/label so we need to dump each individual object

MartinHjelmare:
Ok.

marcelveldt:
also included the list of supported commandClasses in the dump, this fixes #75 

```
"commandClasses": [
            {
              "id": 48,
              "name": "Binary Sensor",
              "version": 2
            },
            {
              "id": 89,
              "name": "Association Group Information",
              "version": 1
            },
            {
              "id": 90,
              "name": "Device Reset Locally",
              "version": 1
            },
            {
              "id": 94,
              "name": "Z-Wave Plus Info",
              "version": 2
            },
            {
              "id": 112,
              "name": "Configuration",
              "version": 1
            },
            {
              "id": 113,
              "name": "Notification",
              "version": 8
            },
            {
              "id": 114,
              "name": "Manufacturer Specific",
              "version": 2
            },
            {
              "id": 128,
              "name": "Battery",
              "version": 1
            },
            {
              "id": 132,
              "name": "Wake Up",
              "version": 2
            },
            {
              "id": 133,
              "name": "Association",
              "version": 2
            },
            {
              "id": 134,
              "name": "Version",
              "version": 2
            }
          ],
```

marcelveldt:
If we miss any values we can add it to the (custom) state object too.

src/lib/state.ts
```diff
@@ -155,6 +167,14 @@ export const dumpDeviceClass = (
   mandatoryControlledCCs: deviceClass.mandatoryControlledCCs,
 });
 
+export const dumpCommandClass = (
+  commandClass: CommandClass
+): CommandClassState => ({
+  id: commandClass.ccId,
+  name: CommandClasses[commandClass.ccId],
+  version: commandClass.version,
```
AlCalzone, line 175: I guess whether this CC is secure or not is of no interest to HA, but it could be useful for debugging.

marcelveldt: > I guess whether this CC is secure or not is of no interest to HA, but it could be useful for debugging.

I can add it. How to access it from the CommandClass instance ?

marcelveldt: nvm, got it

AlCalzone (approved):
The empty interfaces look a bit weird (I'd use plain types for that). Otherwise LGTM

marcelveldt:
> The empty interfaces look a bit weird (I'd use plain types for that). Otherwise LGTM

Yeah, we tend to override/extend from the original interfaces so it will break (intentionally) when upstream changes so we keep up with the changes instead of things silently breaking. But if you have any good ideas to deal with it, let me know!

AlCalzone:
I mean instead of
```ts
interface Foo extends SomeGeneric<Bar> {}
``` 
I'd do
```ts
type Foo = SomeGeneric<Bar>
```

marcelveldt:
@AlCalzone adjusted it.

balloob (approved):

PR #139, opened 2021-02-25
Add basic schema version support

marcelveldt:
Like discussed we're seeking for a way to be compatible between server and HA versions.
I've implemented a very basic api schema for this.

1) client connects --> server sends back version info including the schema versions it can handle:

`
{"type":"version","driverVersion":"6.5.0","serverVersion":"1.0.0","homeId":3967882672,"minSchemaVersion":0,"maxSchemaVersion":1}
`

2) Client decides what to do. For example drop connection if the supported server schema is too old or just handle the supported scheme itself. For example most/all basic commands will just work but relatively new commands won't and the client decides to only not handle the stuff in the upgraded scheme.

3) Client needs to tell the server what schema it wants to use. This is done in the "start_listening" command:

`
{
  "command": "start_listening",
  "messageId": 1,
  "schemaVersion": 1
}
`

From this moment the server knows how to treat commands to/from this client. The server can handle multiple clients with different schema versions.

4) If no schema version parameter is provided in the start_listening command, the server will use the minimal Schema version (which is 0 at this time).

5) If the client sends a schema version which is out of range, this will produce an error to the client an in the server's log:

`
{
  "command": "start_listening",
  "messageId": 1,
  "schemaVersion": 3
}
{"type":"result","success":false,"messageId":1,"errorCode":"schema_incompatible"}
`

6) When we make breaking changes in the api, we should bump the schema version. When adding new commands/features, also bump the api schema and note in both code comments and documentation to which schema version that feature is compatible with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,13 +62,13 @@ interface {
 ```
 
 To start receive the state and get events, the client needs to send the `start_listening` command.
-See below for info about schemeVersion.
+See below for info about schemaVersion.
 
 ```ts
 interface {
   messageId: string;
   command: "start_listening";
-  schemeVersion: 1;
+  schemaVersion: 1;
 }
 ```
 
@@ -113,13 +113,13 @@ interface {
 interface {
   messageId: string;
   command: "start_listening";
-  schemeVersion: 1;
+  schemaVersion: 1;
 }
 ```
 
 #### Update the logging configuration
 
-[compatible with scheme version: 1+]
+[compatible with schema version: 1+]
 
 > NOTE: You must provide at least one key/value pair as part of `config`
 
@@ -139,7 +139,7 @@ interface {
 
 #### Get the logging configuration
 
-[compatible with scheme version: 1+]
+[compatible with schema version: 1+]
 
 ```ts
 interface {
@@ -170,7 +170,7 @@ interface {
 
 #### [Set value on a node](https://zwave-js.github.io/node-zwave-js/#/api/node?id=setvalue)
 
-[compatible with scheme version: 0+]
+[compatible with schema version: 0+]
 
 ```ts
 interface {
@@ -189,7 +189,7 @@ interface {
 
 #### [Refresh node info](https://zwave-js.github.io/node-zwave-js/#/api/node?id=refreshinfo)
 
-[compatible with scheme version: 0+]
+[compatible with schema version: 0+]
 
 ```ts
 interface {
@@ -201,7 +201,7 @@ interface {
 
 #### [Get defined Value IDs](https://zwave-js.github.io/node-zwave-js/#/api/node?id=getdefinedvalueids)
 
-[compatible with scheme version: 0+]
+[compatible with schema version: 0+]
 
 ```ts
 interface {
@@ -213,7 +213,7 @@ interface {
 
 #### [Get value metadata](https://zwave-js.github.io/node-zwave-js/#/api/node?id=getvaluemetadata)
 
-[compatible with scheme version: 0+]
+[compatible with schema version: 0+]
 
 ```ts
 interface {
@@ -231,7 +231,7 @@ interface {
 
 #### [Abort Firmware Update](https://zwave-js.github.io/node-zwave-js/#/api/node?id=abortfirmwareupdate)
 
-[compatible with scheme version: 0+]
+[compatible with schema version: 0+]
 
 ```ts
 interface {
@@ -243,7 +243,7 @@ interface {
 
 #### [Poll value](https://zwave-js.github.io/node-zwave-js/#/api/node?id=pollvalue)
 
-[compatible with scheme version: 1+]
+[compatible with schema version: 1+]
 
 ```ts
 interface {
@@ -261,7 +261,7 @@ interface {
 
 #### [Set raw configuration parameter value (Advanced)](https://zwave-js.github.io/node-zwave-js/#/api/CCs/Configuration?id=set)
 
-[compatible with scheme version: 1+]
+[compatible with schema version: 1+]
 
 ```ts
 interface {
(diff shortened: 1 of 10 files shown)


marcelveldt:
forgot to mention: applying this change will make the server compatible again with current HA stable release.
For HA beta we need to request schema version 1 in the lib...

README.md
```diff
@@ -62,11 +62,13 @@ interface {
 ```
 
 To start receive the state and get events, the client needs to send the `start_listening` command.
+See below for info about schemeVersion.
```
MartinHjelmare, line 65: Should we call it `scheme` or `schema`?

README.md
```diff
@@ -250,6 +271,53 @@ interface {
 }
 ```
 
+## Scheme Version
+
+In an attempt to keep compatibility between different server and client versions, we introduced a (basic) API Scheme Version.
```
MartinHjelmare, line 276:
```suggestion
In an attempt to keep compatibility between different server and client versions, we have introduced a (basic) API Scheme Version.
```

src/lib/forward.ts
```diff
@@ -95,8 +112,14 @@ export class EventForwarder {
 
     node.on("ready", (changedNode: ZWaveNode) => {
       // Dump full node state on ready event
-      const nodeState = dumpNode(changedNode);
-      notifyNode(changedNode, "ready", { nodeState });
+      this.clients.clients.forEach((client) =>
+        this.sendEvent(client, {
```
MartinHjelmare, line 116: We no longer use `notifyNode` here.

MartinHjelmare (commented):
Do we not need to handle the log config get and update methods in a special way? They were added in version 1.

README.md
marcelveldt: I'm not sure...

src/lib/forward.ts
marcelveldt: Correct, is no longer needed. notifyNode was a helper

marcelveldt:
> Do we not need to handle the log config get and update methods in a special way? They were added in version 1.

yep, I've handled them in both the readme and in the lib PR

README.md
marcelveldt: @MartinHjelmare looks like we need to rename to schema:
https://english.stackexchange.com/questions/40702/difference-between-scheme-and-schema

OK ?

marcelveldt: adjusted it to schema

MartinHjelmare:
The general concept looks good! I'd like @balloob to review also.

src/lib/server.ts
```diff
@@ -65,9 +72,23 @@ class Client {
...
 
     try {
+      if (msg.command === DriverCommand.setApiSchema) {
+        // Handle schema version
+        this.schemaVersion = msg.schemaVersion;
+        if (
+          this.schemaVersion < minSchemaVersion ||
+          this.schemaVersion > maxSchemaVersion
+        ) {
+          throw new SchemaIncompatibleError(this.schemaVersion);
```
balloob, line 82: Instead of throwing we should send an error to the client.

marcelveldt: Yes, that is handled

src/lib/forward.ts
```diff
@@ -95,8 +112,14 @@ export class EventForwarder {
...
+          source: "node",
+          event: "ready",
+          nodeId: changedNode.nodeId,
+          ...dumpNode(changedNode, client.schemaVersion),
```
balloob, line 120: You sure about the `...` ? Shouldn't it be `node: dumpNode(…)` ?

balloob: That's how we do it for other times we send over the node. Otherwise we can't distinguish what is event message vs what is the node.

src/lib/server.ts
marcelveldt: Schermafbeelding 2021-02-25 om 20 27 10

balloob (approved):
Looks good. 1 comment.

src/lib/forward.ts
marcelveldt: Yes, this is the additional data

marcelveldt: I think you're right. It was automagically handled by JS but could to explictly return it

balloob: image

All these properties are now added to the event object instead of being under `node`

src/lib/server.ts
balloob: Ok 👍

src/lib/forward.ts
marcelveldt: fixed!

README.md
```diff
@@ -301,7 +300,7 @@ In an attempt to keep compatibility between different server and client versions
 
    From this moment the server knows how to treat commands to/from this client. The server can handle multiple clients with different schema versions.
 
-4. By default the server will use the minimum schema it supports if this command is omitted (which is 0 at this time).
+4. By default the server will use the minimum schema it supports if the `set_api_schema` command is omitted (which is 0 at this time).
```
MartinHjelmare, line 303:
```suggestion
4. By default the server will use the minimum schema it supports (which is 0 at this time) if the `set_api_schema` command is omitted.
```

src/lib/forward.ts
```diff
@@ -95,8 +112,14 @@ export class EventForwarder {
...
     node.on("ready", (changedNode: ZWaveNode) => {
       // Dump full node state on ready event
-      const nodeState = dumpNode(changedNode);
-      notifyNode(changedNode, "ready", { nodeState });
+      this.clients.clients.forEach((client) =>
+        this.sendEvent(client, {
+          source: "node",
+          event: "ready",
+          nodeId: changedNode.nodeId,
+          node: dumpNode(changedNode, client.schemaVersion),
```
balloob, line 120: So for backwards compat, this should be `nodeState`. That's the line that we're replacing.

balloob:
```suggestion
          nodeState: dumpNode(changedNode, client.schemaVersion),
```

raman325: this makes sense to me

balloob (approved):

PR #157, opened 2021-03-12
Create a new schema version to handle switch from string to buffer type for buffer values

raman325:
In this PR, we are bumping the max schema version to 2. For schema version 1 or lower, we change the value type from `buffer` to `string` when found.

TODO:
- [x] Handle changing the value type for "metadata updated" event (`args.metadata.type` needs to be checked and updated)
- [x] Handle changing the value type for "value updated" event (`prevValue.metadata.type` and `newValue.metadata.type` need to be checked and updated)
- [x] Handle changing the value type for "value added" event (`prevValue.metadata.type` needs to be checked and updated)
- [x] Handle changing the value type for "value removed" event (`newValue.metadata.type` needs to be checked and potentially updated)


balloob:
Could we combine this with compatibility with Z-Wave JS Node 7? Or are we going to make that a new schema version?

MartinHjelmare:
Changing logic in an existing schema won't guarantee an effect for users since we can't force an upgrade of the server to a version where those changes have been made.

MartinHjelmare:
But we can't bump the minimum schema version in the client anyway in a patch release of Home Assistant since it would require a server upgrade, which I don't think we should do.

So the least intrusive change is probably still the suggested one and allow users that upgrade the server to get compatibility with zwave-js 6.6.x.

MartinHjelmare:
zwave-js 7.x requires more work to migrate the libraries, so I think it's safer to use separate schema versions for 6.6.x and 7.x to allow us to divide the work. That way we could still bump the required schema version in the client to require the schema we add here for 6.6.x even if we're not done with the work for 7.x.

raman325:
agreed, we should have separate schema versions for 6.6 and 7, there is a significant amount of work to do for 7.

balloob:
Every schema version will need to remain maintained in ZJS Server. So if we can combine more changes in one, that would be great. But ok, if it is more work, let's postpoone.

README.md
```diff
@@ -281,7 +281,7 @@ In an attempt to keep compatibility between different server and client versions
      "serverVersion": "1.0.0",
      "homeId": 3967882672,
      "minSchemaVersion": 0,
-     "maxSchemaVersion": 1
+     "maxSchemaVersion": 2
```
balloob, line 284: I don't think that we need to update this. This is just an example connection string.

I do think that we should add a new `API_SCHEMA.md` with the changes introduced in each schema version.

package.json
```diff
@@ -25,7 +25,7 @@
     "ws": "^7.4.2"
   },
   "peerDependencies": {
-    "zwave-js": "^6.5"
+    "zwave-js": "^6.6"
```
balloob, line 28: We should do this in a different PR so the package-lock is also in that PR. In case we need to revert this PR, we don't want to revert the package upgrade.

README.md
raman325: yes agreed. I started writing this up but I don't know what to put in for the previous schema changes. I can get a doc started and then someone can submit a PR to backfill info for schema 1

package.json
raman325: I can do that, but given that the schema change goes hand in hand with the `zwave-js` bump, wouldn't we want them to be combined? It's ok for the schema change to go out before the version bump, but not ok for the version bump to go out before the schema change

balloob: oh good point.

src/lib/forward.ts
```diff
@@ -110,6 +111,40 @@ export class EventForwarder {
...
+    // Value or Metadata updates require potential schema transforms
+    const notifyNodeOnValueOrMetadataUpdates = (
+      node: ZWaveNode,
+      event: string,
+      args: any
+    ) => {
+      let newArgs: any;
+      this.clients.clients.forEach((client) => {
+        // Copy arguments for each client so transforms don't impact all clients
+        Object.assign(newArgs, args);
```
balloob, line 123:
```suggestion
      this.clients.clients.forEach((client) => {
        // Copy arguments for each client so transforms don't impact all clients
        const newArgs = {...args};
```

src/lib/forward.ts
```diff
@@ -110,6 +111,40 @@ export class EventForwarder {
...
+          );
+        }
+        if ("metadata" in newArgs) {
+          schemaTransformValueMetadata(newArgs.metadata, client.schemaVersion);
+        }
+        this.sendEvent(client, {
+          source: "node",
+          event,
+          nodeId: node.nodeId,
+          ...{ newArgs },
```
balloob, line 143: I don't think this is the same, right? When I look at the old code, it had `{ args }`, which meant that there was a single `args` key that contains the value. That would be equivalent here to:
```suggestion
          { args: newArgs },
```

src/util/metadata_handler.ts
```diff
@@ -0,0 +1,21 @@
+import { ValueID, ValueMetadata, ZWaveNode } from "zwave-js";
+
+export const getTransformedValueMetadata = (
```
balloob, line 3: Should we move these functions to `state.ts`, as that is responsible for serializing the state.

src/lib/forward.ts
raman325: This is causing CI to fail. Does `args` need to be in quotes?

raman325: That didn't work either 🤔

raman325: Oh I think it's because we lost the spread operator

raman325: OK I see, we can avoid creating an object and just pass the param directly

src/lib/state.ts
```diff
@@ -288,3 +290,23 @@ export const dumpState = (
...
+  schemaTransformValueMetadata(metadata, schemaVersion);
+  return metadata;
+};
+
+export const schemaTransformValueMetadata = (
+  metadata: ValueMetadata,
+  schemaVersion: number
+): void => {
+  if (schemaVersion < 2 && metadata.type === "buffer") {
+    metadata.type = "string";
```
balloob, line 310: Here you're actually changing a ZJS type. When we copied args to newArgs, we did a shallow clone, only the top level object has been copied. All values are still pointing at the same instance.

I think that instead of trying to pass classes to JSON serialize, we should instead write out the types that we want like we do for node.

raman325: Not totally following the second part. Are you suggesting we create a `MetadataState` like we have a `NodeState` and `ValueState` and add a `dumpMetadata` function that explicitly copies the properties over while updating `metadata.type` if needed?

raman325: Not sure if I took the right path here but attempted to follow your suggestion in the latest commit.

balloob (approved):

PR #165, opened 2021-03-16
Add support for node-zwave-js 7.0.0 via new server schema 3

raman325:
Reference doc: https://github.com/home-assistant-libs/zwave-js-server-python/issues/162

TODO:

- [x] translate `protocolVersion` into a `version` that would be recognized for schemas < 3
- [x] Handle notification event changes (The code has been updated but it does the same thing for all schema versions, and as mentioned in Discord, I think this will break the existing HA code)


src/lib/controller/message_handler.ts
```diff
@@ -96,7 +98,17 @@ export class ControllerMessageHandler {
...
+          throw new NotSupportedBySchemaError(client.schemaVersion, "<3");
+        }
+        await driver.controller.removeNodeFromAllAssociations(message.nodeId);
+        return {};
+      }
+      case ControllerCommand.removeNodeFromAllAssociations: {
+        if (client.schemaVersion < 3) {
+          throw new NotSupportedBySchemaError(client.schemaVersion, ">2");
+        }
+        await driver.controller.removeNodeFromAllAssociations(message.nodeId);
```
balloob, line 111: Alternatively, we could also just accept both commands and not raise. 
```suggestion
      case ControllerCommand.removeNodeFromAllAssociations: {
        await driver.controller.removeNodeFromAllAssociations(message.nodeId);
```

balloob: (this suggest won't work because of the curly braces, but something like that.

src/lib/state.ts
```diff
@@ -97,19 +97,19 @@ interface NodeStateSchema0 {
...
-  isFrequentListening: ZWaveNode["isFrequentListening"];
+  isFrequentListening: boolean;
   isRouting: ZWaveNode["isRouting"];
-  maxBaudRate: ZWaveNode["maxBaudRate"];
+  maxBaudRate: ZWaveNode["maxDataRate"];
   isSecure: ZWaveNode["isSecure"];
-  version: ZWaveNode["version"];
-  isBeaming: ZWaveNode["isBeaming"];
+  version?: number;
+  isBeaming: ZWaveNode["supportsBeaming"];
```
balloob, line 105: If we do this, we should make sure we require at least version 7 as peer dependency and we should mark that in our readme too.

src/lib/state.ts
```diff
@@ -241,19 +267,19 @@ export const dumpNode = (
     status: node.status,
     ready: node.ready,
     isListening: node.isListening,
-    isFrequentListening: node.isFrequentListening,
+    isFrequentListening: Boolean(node.isFrequentListening),
```
balloob, line 265: Why wrap it in boolean, if it's undefined, we should put in `null` to indicate we don't know

src/lib/state.ts
```diff
@@ -241,19 +267,19 @@ export const dumpNode = (
...
     isRouting: node.isRouting,
-    maxBaudRate: node.maxBaudRate,
+    maxBaudRate: node.maxDataRate,
```
balloob, line 267: Don't add this here since we're going to remove it in schema 3. Only here add the base that's the same across all versions.

src/lib/state.ts
raman325: The type used to be a `boolean`, I think we always returned either `true` or `false`

src/lib/state.ts
raman325: the `version` property here references the protocol version that a node supports. I think you are talking about `node-zwave-js` right?

src/lib/state.ts
raman325: You were right. Updated

src/lib/controller/incoming_message.ts
```diff
@@ -96,12 +96,20 @@ export interface IncomingCommandControllerRemoveAssociations
...
   command: ControllerCommand.removeNodeFromAllAssocations;
   nodeId: number;
 }
 
+// Schema version > 2
+export interface IncomingCommandControllerRemoveNodeFromAllAssociations
+  extends IncomingCommandControllerBase {
+  command: ControllerCommand.removeNodeFromAllAssociations;
+  nodeId: number;
+}
```
balloob, line 111: Do we want to specify 2 types or do we want to specify 1 interface that represents both commands ?

package.json
```diff
@@ -41,7 +42,7 @@
     "prettier": "^2.2.1",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.3",
-    "zwave-js": "^6.6"
+    "zwave-js": "^7.0.0-beta.0"
```
balloob, line 45: Should be beta 1 too ?

src/lib/server.ts
```diff
@@ -106,6 +107,16 @@ export class Client {
       if (msg.command === DriverCommand.getLogConfig) {
         // We don't want to return transports since that's used internally.
         const { transports, ...partialLogConfig } = this.driver.getLogConfig();
+
+        if (
+          this.schemaVersion < 3 &&
+          typeof partialLogConfig.level === "string"
+        ) {
+          let levelNum = numberFromLogLevel(partialLogConfig.level);
+          if (levelNum != undefined) {
```
balloob, line 116: Should we make it an error when it is undefined ?

balloob: I guess it's the old schema, yolo.

src/lib/state.ts
balloob: I meant that the types don't work when we install z-wavejs 6, but I guess that doesn't matter because they fail anyway in that case.

balloob: Actually, we should probably write out ALL types (in a new PR), so it's easy to spot type errors when we update versions, so we know what needs to be made compatible.

balloob (approved):

package.json
raman325: thanks missed that one

src/lib/server.ts
raman325: I suppose it wouldn't hurt as a best practices kinda thing

raman325: oh that's right, this is coming from node-zwave-js, so I don't think we need to worry about it being undefined

src/lib/state.ts
raman325: oh I see, yeah we can do that in a separate PR

PR #167, opened 2021-03-18
Fix notification event, add missing properties, and update client to accept schemaVersion

raman325:
I made two mistakes:

- Referenced the wrong argument in the notification event
- I was missing properties in schema 3

This PR corrects it.

I also added a `schemaVersion` argument for the client and default to using the latest schema (we currently always use schema 0 which makes testing a little hard 😉 )

### DIFF
--- a/src/bin/client.ts
+++ b/src/bin/client.ts
@@ -19,11 +19,8 @@ const schemaVersion = args.schemaVersion
 const url = args._[0] || "ws://localhost:3000";
 const filterNode = args.node ? Number(args.node) : undefined;
 
-if (schemaVersion > maxSchemaVersion) {
-  console.log(
-    "Schema version must be less than or equal to ",
-    maxSchemaVersion
-  );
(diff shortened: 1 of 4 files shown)


src/bin/client.ts
```diff
@@ -1,26 +1,43 @@
...
 
-const args = parseArgs<Args>(["_", "dump", "node"]);
+const args = parseArgs<Args>(["_", "dump", "node", "schemaVersion"]);
+const schemaVersion = args.schemaVersion
+  ? Number(args.schemaVersion)
+  : maxSchemaVersion;
 const url = args._[0] || "ws://localhost:3000";
 const filterNode = args.node ? Number(args.node) : undefined;
 
+if (schemaVersion > maxSchemaVersion || schemaVersion < 0) {
```
balloob, line 22:
```suggestion
if (isNaN(schemaVersion) || schemaVersion > maxSchemaVersion || schemaVersion < 0) {
```

balloob (approved):

PR #168, opened 2021-03-19
Bump zwave-js to 7.0.0-beta2

raman325:
As part of updating to `7.0.0-beta2` the endpoint schema has changed. In this PR we add the new deviceClass property while considering backwards compatibility

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -108,7 +108,7 @@ interface NodeStateSchema0 {
   isRouting: ZWaveNode["isRouting"];
   maxBaudRate: ZWaveNode["maxDataRate"];
   isSecure: ZWaveNode["isSecure"];
-  version?: number;
+  version: number | null;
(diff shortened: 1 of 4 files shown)


src/lib/state.ts
```diff
@@ -341,12 +346,25 @@ export const dumpNode = (node: ZWaveNode, schemaVersion: number): NodeState => {
...
+  let base: EndpointStateSchema0 = {
+    nodeId: endpoint.nodeId,
+    index: endpoint.index,
+    installerIcon: endpoint.installerIcon,
+    userIcon: endpoint.userIcon,
+  };
+  if (schemaVersion < 3) {
+    return base as EndpointStateSchema0;
+  }
+  const endpoint3 = base as EndpointStateSchema1;
```
balloob, line 362: Since we're still changing the v3 schema, we should mark it as beta.

balloob (approved):

PR #170, opened 2021-03-24
Bump zwave-js-server to 1.3.0

raman325:
This should follow https://github.com/zwave-js/zwave-js-server/pull/169

Also would we consider this a minor release to support zwave-js 7.0.0 or should we bump to 2.0.0?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


MartinHjelmare:
Probably bump major version if we make breaking changes.

raman325:
> Probably bump major version if we make breaking changes.

The changes are intended to be backwards compatible so there are no breaking changes in theory. I tested it for a bit and the backwards compatibility seemed to be working.

balloob:
We cannot release v2.0.0 without breaking the 2021.1 release. Also, we're aiming for it to be backwards compatible, so 1.3.0 is good.

raman325:
> We cannot release v2.0.0 without breaking the 2021.1 release. Also, we're aiming for it to be backwards compatible, so 1.3.0 is good.

at what point are we ok breaking the 2021.1 release? I am hoping this does not mean we can't ever reach 2.0.0 🙂

PR #177, opened 2021-03-29
Dump Configuration CC Metadata

raman325:
I missed adding metadata properties that are specific to the Configuration CC

Fixes https://github.com/home-assistant/core/issues/48483

Once this PR is merged, I think we can drop the `as ConfigurationMetadata` part: https://github.com/zwave-js/node-zwave-js/pull/2193

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -99,6 +99,19 @@ interface MetadataState {
   steps?: number;
   states?: Record<number, string>;
   unit?: string;
+}
+
+interface ConfigurationMetadataState {
+  type: ConfigurationMetadata["type"];
+  readable: boolean;
+  writeable: boolean;
+  description?: string;
+  label?: string;
+  ccSpecific?: Record<string, any>;
+  min?: ConfigValue;
+  max?: ConfigValue;
+  default?: ConfigValue;
+  unit?: string;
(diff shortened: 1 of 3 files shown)


src/lib/state.ts
```diff
@@ -81,21 +86,44 @@ interface ValueState extends TranslatedValueID {
 
 interface MetadataState {
```
balloob, line 87: Should this be renamed to better distinguish it from `ConfigurationMetadataState` ?

balloob (approved):

src/lib/state.ts
raman325: suggestions for what to call it? Not a bad idea, just don't know a better name 🤷🏾

raman325:
Let's hold on merging until https://github.com/zwave-js/node-zwave-js/pull/2193 is merged and released since it will change where we import from and simplify some of the code.

raman325:
> Let's hold on merging until [zwave-js/node-zwave-js#2193](https://github.com/zwave-js/node-zwave-js/pull/2193) is merged and released since it will change where we import from and simplify some of the code.

Actually we need to merge this - the set config parameter services will not work correctly for cases where `allowManualEntry` is true without this change

PR #180, opened 2021-03-31
Handle upstream configuration CC value metadata changes

raman325:
See zwave-js/node-zwave-js#2193 for more context. This PR will need to be merged and released alongside the release of that PR so that things don't break in `zwavejs2mqtt`. We will also need to update `zwave-js` and `zwave-js-server` at the same time in the addon

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -158,63 +158,63 @@
       }
     },
     "@sentry/core": {
-      "version": "6.2.3",
-      "resolved": "https://registry.npmjs.org/@sentry/core/-/core-6.2.3.tgz",
-      "integrity": "sha512-GpfHoSJiXchVXgyaMWVtIPVw2t97KkD1OJ4JdL3/TeH3auX5XvsN5iHTk+x/Er8t13IpOnvidH1xWdV1dnax2w==",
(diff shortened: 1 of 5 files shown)


raman325:
@balloob I added some logic to check types for `min` and `max` to get around some typing errors, just wanted a second pair of eyes on it in case there was a better way to do this. The issue is that because the `ValueMetadata` type includes the `ConfigurationValue` type, typescript assumes that `min` and `max` can be of type `ConfigValue` which is only for the Configuration CC and is a superset of the types that are supported in other CCs

raman325:
The only way I've found to make this work better is to explicitly Union all of the metadata types. I know we had talked about being more explicit about typing, do we want to do that here too? There's an increased maintenance cost with that approach

raman325:
> The only way I've found to make this work better is to explicitly Union all of the metadata types. I know we had talked about being more explicit about typing, do we want to do that here too? There's an increased maintenance cost with that approach

Decided to be explicit about types.

raman325:
Going to go ahead and merge this to unblock the other PRs and because I think this is as clean as I can make it. I can adjust if needed

PR #181, opened 2021-04-02
Add support for new events

raman325:
Two new events were added in `zwave-js 7.1.0`
- `interview started`
- `interview stage completed`

We want to forward these events on so that we can track interview progress when calling `node.refresh_info` from the python lib.

~~Al does not want a `not ready` event in zwave-js so we also add one here so that the lib can update ready status in coordination with the driver~~  EDIT: We can use the `interview started` event to tell the library to set ready back to false


MartinHjelmare:
What does the `"not ready"` event mean?

raman325:
It means that ready has switched from `true` to `false`. It only occurs on startup if a full interview is required (in which case we wouldn't fire the event) and when `node.refreshInfo` is called (in which case we would fire the event)

MartinHjelmare:
There was no further discussion how a node that goes from ready to not ready is supposed to be handled in zwave-js?

raman325:
> There was no further discussion how a node that goes from ready to not ready is supposed to be handled in zwave-js?

In terms of? Maybe I am not understanding but internally, `zwave-js` changes the `node.ready` property as needed, we just don't get an event for it so the library doesn't update its internal state without this new event. Per this comment https://github.com/zwave-js/node-zwave-js/pull/2233#pullrequestreview-626896636 the suggestion was to fire the event from server directly

MartinHjelmare:
Right. Since zwave-js expose the ready property and ready event of the node I'd assume there should be a strategy for library users when the node is not ready anymore, whatever is the reason for that.

raman325:
My assumption so far is that when a node is not ready, it should not be interacted with. That would mean making the entities unavailable during the re-interview process. I've asked Al if he agrees or has any other guidance

balloob:
Don't we have some `node updated` evnet we can used to propagate ready update? Feels a bit dirty to invent our own. If it's not needed inside ZJS, why do we need it.

raman325:
> Don't we have some `node updated` evnet we can used to propagate ready update? Feels a bit dirty to invent our own. If it's not needed inside ZJS, why do we need it.

Well we have to emit the event. We could co-opt the `node updated` event. I believe the library code allows us to send partial info, e.g. just send the new ready state

raman325:
@MartinHjelmare see Al's response here: https://github.com/zwave-js/node-zwave-js/pull/2233#issuecomment-813114569

So the right thing to do from a Home Assistant perspective is to mark the entities as unavailable. This logic is already correct, what we need to do is update the ready status in the library and then react to the event in the integration.

raman325:
> Don't we have some `node updated` event we can used to propagate ready update? Feels a bit dirty to invent our own. If it's not needed inside ZJS, why do we need it.

OK so I did some testing, and we can just use the `interview started` event to reset the cached `ready` state in the library. Removing the extra event. Here's what it looks like now btw (before removing the extra event):
```json
{
  type: 'event',
  event: { source: 'node', event: 'interview started', nodeId: 4 }
}
{
  type: 'event',
  event: { source: 'node', event: 'not ready', nodeId: 4 }
}
{
  type: 'result',
  success: true,
  messageId: 'refresh-info',
  result: {}
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'ProtocolInfo'
  }
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'NodeInfo'
  }
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'CommandClasses'
  }
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'OverwriteConfig'
  }
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'Neighbors'
  }
}
{
  type: 'event',
  event: {
    source: 'node',
    event: 'interview stage completed',
    nodeId: 4,
    stageName: 'Complete'
  }
}
```

PR #186, opened 2021-04-06
Split driver commands out into separate handler and add new commands

raman325:
The existing `get_log_config` and `update_log_config` commands are now available as `driver.get_log_config` and `driver.update_log_config`. In addition:
- `driver.enable_statistics`, `driver.disable_statistics`, `driver.statistics_enabled`, and `driver.get_config` commands added
- `dumpState` now include the driver config

### DIFF
--- a/src/lib/driver/command.ts
+++ b/src/lib/driver/command.ts
@@ -1,0 +1,6 @@
+export enum DriverCommand {
+  updateLogConfig = "driver.update_log_config",
+  getLogConfig = "driver.get_log_config",
+  enableStatistics = "driver.enable_statistics",
+  disableStatistics = "driver.disable_statistics",
+}
--- a/src/lib/driver/command.ts
+++ b/src/lib/driver/command.ts
@@ -4,5 +4,5 @@ export enum DriverCommand {
   getLogConfig = "driver.get_log_config",
   enableStatistics = "driver.enable_statistics",
   disableStatistics = "driver.disable_statistics",
-  statisticsEnabled = "driver.statistics_enabled",
+  isStatisticsEnabled = "driver.is_statistics_enabled",
 }
--- a/src/lib/driver/command.ts
+++ b/src/lib/driver/command.ts
@@ -1,6 +1,8 @@
 export enum DriverCommand {
+  getConfig = "driver.get_config",
   updateLogConfig = "driver.update_log_config",
   getLogConfig = "driver.get_log_config",
   enableStatistics = "driver.enable_statistics",
   disableStatistics = "driver.disable_statistics",
+  statisticsEnabled = "driver.statistics_enabled",
 }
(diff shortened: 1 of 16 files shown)


MartinHjelmare:
To me it sounds good to split driver commands out. We have separate folders for controller and node already, so it's in line with our existing design I think.

src/lib/driver/command.ts
```diff
@@ -0,0 +1,6 @@
+export enum DriverCommand {
+  updateLogConfig = "driver.update_log_config",
+  getLogConfig = "driver.get_log_config",
```
balloob, line 3: Would be nice to create some generic "getConfig" command to get all driver info at once.

src/lib/driver/command.ts
```diff
@@ -0,0 +1,6 @@
...
+  enableStatistics = "driver.enable_statistics",
+  disableStatistics = "driver.disable_statistics",
```
balloob, line 5: Do we want to have this as 2 methods, or 1 `set_statistics` method or even have 1 `set_config` method ?

src/lib/driver/command.ts
raman325: What other information do you want to get about the driver? Right now we don't model the state of the driver at all in the library

src/lib/driver/command.ts
raman325: I don't think we gain much by having a single `set_config` method because you will set different parts of the config at different times. Changing the log level will be random, enabling statistics will be when the client connects to the server if the user opted in or when the user opts in, disabling statistics will be when the user opts out.

As to one command or two, I don't have a preference other than the fact that the commands have a clean relationship with the API methods in the `zwave-js` docs if they remain separate

balloob: following ZJS Methods makes sense, since we follow their structure too.

src/lib/driver/command.ts
balloob: We can start with what we have:

```ts
interface Driver {
  logConfig: LogConfig;
  statisticsEnabled: boolean;
}
```

balloob: (ofc we should mimic the names of the properties on the `driver` object)

raman325: When we have only one client, this can be done easily because we can update the driver state when we enabled/disable statistics or update the log config after getting the initial state. When we have multiple clients, maintaining this state down to the lib would require new events to be passed to all clients whenever those commands are called.

Do we want to maintain this state down to the library, or is the goal here just to maintain the state of the driver in the server for now?

balloob: Let's not add things to the lib unless we are going to use it.

src/lib/driver/command.ts
```diff
@@ -1,6 +1,8 @@
 export enum DriverCommand {
+  getConfig = "driver.get_config",
   updateLogConfig = "driver.update_log_config",
   getLogConfig = "driver.get_log_config",
   enableStatistics = "driver.enable_statistics",
   disableStatistics = "driver.disable_statistics",
+  statisticsEnabled = "driver.statistics_enabled",
```
balloob, line 7: can you add a word to describe what is done with `statistics_enabled`, maybe `is_statistics_enabled` ?

balloob (approved):

PR #188, opened 2021-04-10
Add node refresh_values and refresh_cc_values commands

kpine:
Bump API schema to 4 to account for new commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,32 @@ interface {
 }
 ```
 
+#### [Refresh values](https://zwave-js.github.io/node-zwave-js/#/api/node?id=refreshvalues)
+
+[compatible with schema version: 4+]
+
+```ts
+interface {
+  messageId: string;
+  command: "node.refresh_values";
+  nodeId: number;
+}
+```
+
+#### [Refresh command class values](https://zwave-js.github.io/node-zwave-js/#/api/node?id=refreshccvalues)
+
+[compatible with schema version: 4+]
+
+```ts
+interface {
+  messageId: string;
+  command: "node.refresh_cc_values";
+  nodeId: number;
+  commandClass: CommandClasses;
+}
+```
+
+
 ## Schema Version
 
 In an attempt to keep compatibility between different server and client versions, we've introduced a (basic) API Schema Version.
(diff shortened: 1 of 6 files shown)


raman325:
We shouldn't need to bump the schema for this - the schema is more about when the structure of the data we receives changes and we need to account for older clients (e.g. the current version of zwave-js-server-python that's being used in HA)

kpine:
I didn't think it was necessary, but the README says so?

> When we make breaking changes in the api, we bump the schema version. **When adding new commands/features, we also bump the api schema and note in both code comments and documentation to which schema version that feature is compatible with.**

I can revert it if that is not true.

raman325:
Ah I wasn't aware of that. We do have to bump the schema to 4 anyway I think for the next release so let's leave it as is for now

kpine:
Sure, sounds good, thanks.

If you're an updated application wanting to use these commands and you're talking to an older server, without a new schema to check against would you just handle the error thrown by the older server? There's no schema check anyways for incoming node messages.

raman325:
Yeah, we have a catch all to catch errors and return a failure.

If a user is using zwavejs2mqtt, the version of zwave-js gets bumped alongside bumps to zwave-js-server such that you wouldn't have such a version mismatch. If you were to do a Docker build of the server, same thing because we pin the requirements in package.json. The official addon pins the two separately but they're also coordinated in lockstep.

That's why, to me, we only really need to bump schema versions when we change the data model of a command and not when we add a new one, but @marcelveldt wrote the schema logic and wrote that up in the docs so I wouldn't want to diverge from it until we got more input

balloob:
Increasing the version schema won't break the old schema version, it's still supported. It's just to make sure that we will never expect that method if the client is in schema 3.

raman325:
> Increasing the version schema won't break the old schema version, it's still supported. It's just to make sure that we will never expect that method if the client is in schema 3.

So then should my other PR (https://github.com/zwave-js/zwave-js-server/pull/186) be schema 4 as well?

raman325:
> > Increasing the version schema won't break the old schema version, it's still supported. It's just to make sure that we will never expect that method if the client is in schema 3.
> 
> So then should my other PR (#186) be schema 4 as well?

Went ahead and listed them as schema 4

raman325 (approved):
looks good to me

PR #194, opened 2021-04-13
Make interviewStage property a string (schema 4)

raman325:
Right now we store the enum number value which is subject to change according to Al. It will be more useful to get the string translation of the interview stage. We shouldn't depend on the string value either but this way we don't have to maintain a mapping downstream

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -186,19 +186,22 @@ type NodeStateSchema1 = Modify<
 
 type NodeStateSchema2 = NodeStateSchema1;
 
-type NodeStateSchema3 = Modify<
-  Omit<NodeStateSchema2, "maxBaudRate" | "version" | "isBeaming" | "roleType">,
-  {
-    isFrequentListening: ZWaveNode["isFrequentListening"];
-    maxDataRate: ZWaveNode["maxDataRate"];
-    supportedDataRates: ZWaveNode["supportedDataRates"];
-    protocolVersion: ZWaveNode["protocolVersion"];
-    supportsBeaming: ZWaveNode["supportsBeaming"];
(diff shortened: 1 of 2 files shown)


raman325:
I couldn't figure out how to avoid having to set all of those extra properties for schema 3 and 4 separately because I was getting type errors. Open to suggestions on how to overcome that

src/lib/state.ts
```diff
@@ -380,13 +384,14 @@ export const dumpNode = (node: ZWaveNode, schemaVersion: number): NodeState => {
...
     aggregatedEndpointCount: node.aggregatedEndpointCount,
     interviewAttempts: node.interviewAttempts,
-    interviewStage: node.interviewStage,
     endpoints: Array.from(node.getAllEndpoints(), (endpoint) =>
       dumpEndpoint(endpoint, schemaVersion)
     ),
     values: getNodeValues(node, schemaVersion),
   };
 
+  if (schemaVersion <= 3) base.interviewStage = node.interviewStage;
```
balloob, line 393: you can put this in the existing if-statement below?

balloob:
Yeah working with multiple data formats, TypeScript is not our friend :) it's ok to keep them duplicate.

src/lib/state.ts
raman325: The statement below is `< 3`, this is `<=3`

balloob: oh :)

raman325:
@balloob any additional feedback on this or https://github.com/zwave-js/zwave-js-server/pull/181? I'd like to get these merged soon so we can add support for them to the library and unblock Charles for the frontend work

balloob (approved):
Oops, Forgot to hit approve.

PR #195, opened 2021-04-16
Add logging forwarder

raman325:
Two server level commands have been added:

- `start_listening_to_logs` - client will start getting forwarded logs
- `stop_listening_to_logs` - client will stop getting forwarded logs

~~There appears to be an issue where the first time we attach our transport, the transport does not get all the logs. So for now, I've implemented a workaround where we attach and detach our transport immediately. The first time a client wants to listen to logs, because we've already attached once, the second time we attach everything works. I can remove the workaround if/when this issue gets resolved~~ This was resolved by setting the level for the transport.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "ws": "^7.4.2"
   },
   "peerDependencies": {
-    "zwave-js": "^7.2.1"
+    "zwave-js": "^7.2.4"
   },
   "devDependencies": {
     "@types/minimist": "^1.2.1",
@@ -42,7 +42,7 @@
     "prettier": "^2.2.1",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.3",
-    "zwave-js": "^7.2.1"
+    "zwave-js": "^7.2.4"
   },
   "husky": {
     "hooks": {
--- a/src/lib/driver/outgoing_message.ts
+++ b/src/lib/driver/outgoing_message.ts
@@ -9,4 +9,6 @@ export interface DriverResultTypes {
   [DriverCommand.disableStatistics]: Record<string, never>;
   [DriverCommand.enableStatistics]: Record<string, never>;
   [DriverCommand.isStatisticsEnabled]: { statisticsEnabled: boolean };
+  [DriverCommand.startListeningToLogs]: Record<string, never>;
+  [DriverCommand.stopListeningToLogs]: Record<string, never>;
 }
--- a/src/lib/driver/outgoing_message.ts
+++ b/src/lib/driver/outgoing_message.ts
@@ -9,6 +9,6 @@ export interface DriverResultTypes {
   [DriverCommand.disableStatistics]: Record<string, never>;
   [DriverCommand.enableStatistics]: Record<string, never>;
   [DriverCommand.isStatisticsEnabled]: { statisticsEnabled: boolean };
-  [DriverCommand.startListeningToLogs]: Record<string, never>;
-  [DriverCommand.stopListeningToLogs]: Record<string, never>;
+  [DriverCommand.startListeningLogs]: Record<string, never>;
+  [DriverCommand.stopListeningLogs]: Record<string, never>;
 }
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -5,14 +5,10 @@ import {
   ZWaveNodeEvents,
   ZWaveNodeMetadataUpdatedArgs,
 } from "zwave-js";
-import {
-  CommandClasses,
-  ConfigurationMetadata,
-  ZWaveLogInfo,
-} from "@zwave-js/core";
+import { CommandClasses, ConfigurationMetadata } from "@zwave-js/core";
 import { OutgoingEvent } from "./outgoing_message";
 import { dumpConfigurationMetadata, dumpMetadata, dumpNode } from "./state";
-import { Client, ClientsController, ZwavejsServer } from "./server";
+import { Client, ClientsController } from "./server";
 
 export class EventForwarder {
   /**
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -16,7 +16,7 @@ export class EventForwarder {
    *
    * @param clients
    */
-  constructor(public clients: ClientsController) {}
+  constructor(private clients: ClientsController) {}
 
   start() {
     this.clients.driver.controller.nodes.forEach((node) =>
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -30,11 +30,13 @@ export class EventForwarder {
       this.setupNode(node)
     );
 
+    // Forward logging events on to clients that are currently
+    // receiving logs
     this.server.on("logging", (info: ZWaveLogInfo) => {
       this.clients.clients
-        .filter((cl) => cl.receiveLogs)
+        .filter((cl) => cl.receiveLogs && cl.isConnected)
         .forEach((client) =>
-          this.sendEvent(client, {
+          client.sendEvent({
             source: "driver",
             event: "logging",
             info,
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -32,14 +32,14 @@ export class EventForwarder {
 
     // Forward logging events on to clients that are currently
     // receiving logs
-    this.server.on("logging", (info: ZWaveLogInfo) => {
+    this.server.on("logging", (message: string) => {
       this.clients.clients
         .filter((cl) => cl.receiveLogs && cl.isConnected)
         .forEach((client) =>
           client.sendEvent({
             source: "driver",
             event: "logging",
-            info,
+            message,
           })
         );
     });
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -5,10 +5,14 @@ import {
   ZWaveNodeEvents,
   ZWaveNodeMetadataUpdatedArgs,
 } from "zwave-js";
-import { CommandClasses, ConfigurationMetadata } from "@zwave-js/core";
+import {
+  CommandClasses,
+  ConfigurationMetadata,
+  ZWaveLogInfo,
+} from "@zwave-js/core";
 import { OutgoingEvent } from "./outgoing_message";
 import { dumpConfigurationMetadata, dumpMetadata, dumpNode } from "./state";
-import { Client, ClientsController } from "./server";
+import { Client, ClientsController, ZwavejsServer } from "./server";
 
 export class EventForwarder {
   /**
@@ -16,12 +20,27 @@ export class EventForwarder {
    *
    * @param clients
    */
-  constructor(public clients: ClientsController) {}
+  constructor(
+    public clients: ClientsController,
+    private server: ZwavejsServer
+  ) {}
 
   start() {
     this.clients.driver.controller.nodes.forEach((node) =>
       this.setupNode(node)
     );
+
+    this.server.on("logging", (info: ZWaveLogInfo) => {
+      this.clients.clients
+        .filter((cl) => cl.receiveLogs)
+        .forEach((client) =>
+          this.sendEvent(client, {
+            source: "driver",
+            event: "logging",
+            info,
+          })
+        );
+    });
 
     // Bind to all controller events
     // https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/controller/Controller.ts#L112
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -20,29 +20,12 @@ export class EventForwarder {
    *
    * @param clients
    */
-  constructor(
-    public clients: ClientsController,
-    private server: ZwavejsServer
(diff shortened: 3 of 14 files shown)


src/lib/server.ts
```diff
@@ -323,3 +352,14 @@ export class ZwavejsServer extends EventEmitter {
...
   }
 }
+
+export class EventEmitterLogTransport extends Transport {
+  public constructor(private emitter: EventEmitter) {
+    super({ format: createDefaultTransportFormat(false, false) });
+  }
+
+  public log(info: ZWaveLogInfo, next: () => void): any {
+    this.emitter.emit("logging", info);
```
AlCalzone, line 362: You might get away with just emitting `info.message` here. The default transport format should already have formatted it. (Please test though :) )

src/lib/server.ts
```diff
@@ -291,9 +313,16 @@ export class ZwavejsServer extends EventEmitter {
     if (!this.driver.ready) {
       throw new Error("Cannot start server when driver not ready");
     }
+
+    // Create log transport for server and update the drivers log configuration
+    this.serverTransport = new EventEmitterLogTransport(this);
+    const transports = this.driver.getLogConfig().transports;
+    transports.push(this.serverTransport);
+    this.driver.updateLogConfig({ transports });
```
AlCalzone, line 321: Is it possible that this is called multiple times? I guess not, but if it is, you should probably check if the transport was already registered.

src/lib/forward.ts
```diff
@@ -5,24 +5,45 @@ import {
...
+  ) {}
 
   start() {
     this.clients.driver.controller.nodes.forEach((node) =>
       this.setupNode(node)
     );
 
+    // Forward logging events on to clients that are currently
+    // receiving logs
+    this.server.on("logging", (info: ZWaveLogInfo) => {
```
AlCalzone, line 35: Depending what you decide below (emit message or entire info), this could also be `message: string`

src/lib/server.ts
raman325: It shouldn't be possible the way the setup is done: https://github.com/zwave-js/zwave-js-server/blob/master/src/bin/server.ts#L69-L77

src/lib/forward.ts
raman325: I'm going to stick with this for now so that I can see the raw data come in locally and will adjust based on what makes sense

src/lib/server.ts
```diff
@@ -354,8 +354,11 @@ export class ZwavejsServer extends EventEmitter {
 }
 
 export class EventEmitterLogTransport extends Transport {
+  private messageSymbol: Symbol;
+
   public constructor(private emitter: EventEmitter, private logger: Logger) {
     super({ format: createDefaultTransportFormat(false, false) });
+    this.messageSymbol = Symbol.for("message");
```
AlCalzone, line 361: Are you actually using this?

src/lib/forward.ts
```diff
@@ -5,24 +5,45 @@ import {
...
+  ) {}
 
   start() {
     this.clients.driver.controller.nodes.forEach((node) =>
       this.setupNode(node)
     );
 
+    // Forward logging events on to clients that are currently
+    // receiving logs
+    this.server.on("logging", (message: string) => {
```
balloob, line 35: We should attach this only when people are listening, and remove it when last listener is gone. That way we don't create a bunch of overhead.

raman325: Not sure yet how to instrument this on the listener side but will think it through. Is the overhead on the listener side or on the emitter side? Or is the concern that both are always up with this approach?

src/lib/server.ts
raman325: thanks I think I lost something on rebase

src/lib/logging.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(
+    private clients: ClientsController,
+    private server: ZwavejsServer
+  ) {}
+
+  // Forward logging events on to clients that are currently
+  // receiving logs
+  forwardLogging(message: string) {
+    this.clients.clients
```
raman325, line 20: I keep getting the following error here and I can't figure out why:

TypeError: Cannot read property 'clients' of undefined
    at ZwavejsServer.forwardLogging (/usr/src/app/node_modules/@zwave-js/server/dist/lib/logging.js:19:22)
    at ZwavejsServer.emit (node:events:378:20)
    at ZwavejsServer.EventEmitter.emit (node:domain:470:12)
    at EventEmitterLogTransport.log (/usr/src/app/node_modules/@zwave-js/server/dist/lib/server.js:296:22)
    at EventEmitterLogTransport._write (/usr/src/app/node_modules/winston-transport/index.js:103:17)
    at doWrite (/usr/src/app/node_modules/readable-stream/lib/_stream_writable.js:428:64)
    at writeOrBuffer (/usr/src/app/node_modules/readable-stream/lib/_stream_writable.js:417:5)
    at EventEmitterLogTransport.Writable.write (/usr/src/app/node_modules/readable-stream/lib/_stream_writable.js:334:11)
    at DerivedLogger.ondata (/usr/src/app/node_modules/winston/node_modules/readable-stream/lib/_stream_readable.js:681:20)
    at DerivedLogger.emit (node:events:390:22)

raman325: I fixed this by inlining the function instead of having a separate one. The only potential problem with that is that the only way to remove the listener is to remove all listeners. Right now, this isn't a problem but it could be in the future if we decide to listen to the events elsewhere

raman325:
OK I've reworked this so that the LogTransport does the work of forwarding the events to clients. We only add the transport when needed to minimize the overhead that's added.

raman325:
OK so there appears to be an issue where the first time we attach our transport, the transport does not get all the logs. So for now, I've implemented a workaround where we attach and detach our transport immediately. The first time a client wants to listen to logs, because we've already attached once, the second time we attach everything works. I can remove the workaround if/when this issue gets resolved

package.json
```diff
@@ -25,7 +25,7 @@
     "ws": "^7.4.2"
   },
   "peerDependencies": {
-    "zwave-js": "^7.2.1"
+    "zwave-js": "^7.2.4"
```
balloob, line 28: You should do this in a different PR, because dependabot actually does this.

src/lib/driver/outgoing_message.ts
```diff
@@ -9,4 +9,6 @@ export interface DriverResultTypes {
   [DriverCommand.disableStatistics]: Record<string, never>;
   [DriverCommand.enableStatistics]: Record<string, never>;
   [DriverCommand.isStatisticsEnabled]: { statisticsEnabled: boolean };
+  [DriverCommand.startListeningToLogs]: Record<string, never>;
```
balloob, line 12: Should we drop the `To` ?

src/lib/logging.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class WebSocketLogTransport extends Transport {
+  private messageSymbol: Symbol;
+
+  public constructor(private clients: ClientsController) {
+    super({
+      format: createDefaultTransportFormat(false, false),
+    });
+    this.messageSymbol = Symbol.for("message");
```
balloob, line 51: How can this symbol work. The idea of symbols is that they are unique, or do I misunderstand?

src/lib/logging.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Forward logs on to clients that are currently
+    // receiving logs
+    this.clients.clients
+      .filter((cl) => cl.receiveLogs && cl.isConnected)
+      .forEach((client) =>
+        client.sendEvent({
+          source: "driver",
+          event: "logging",
+          formattedMessage: info[this.messageSymbol as any],
+          ...info,
```
balloob, line 64: Will this not include `this.messageSymbol` ?

src/lib/server.ts
```diff
@@ -233,6 +238,36 @@ export class ClientsController {
...
+
+  public configureLoggingEventForwarder() {
+    if (!this.loggingEventForwarderStarted) {
+      this.loggingEventForwarder?.start();
+    }
+  }
+
+  public cleanupLoggingEventForwarder() {
+    if (
+      this.clients.filter((cl) => cl.receiveLogs && cl.isConnected).length ==
```
balloob, line 265: We don't care about connected here I think. 
```suggestion
      this.clients.filter((cl) => cl.receiveLogs).length ==
```

src/lib/server.ts
```diff
@@ -233,6 +238,36 @@ export class ClientsController {
       this.eventForwarder = new EventForwarder(this);
       this.eventForwarder.start();
     }
+
+    if (this.loggingEventForwarder === undefined) {
+      this.loggingEventForwarder = new LoggingEventForwarder(this, this.driver);
+    }
```
balloob, line 244: No need to initialize it until we start it.

balloob: (I know you're following the event forwarding stuff, but that is probably also up for a refactor now that listening is also opt-in and people might just check the supported versions)

src/lib/server.ts
```diff
@@ -233,6 +238,36 @@ export class ClientsController {
...
+  }
+
+  get loggingEventForwarderStarted(): boolean {
+    if (
+      !this.loggingEventForwarder ||
+      this.loggingEventForwarder === undefined
+    ) {
+      return false;
+    }
+    return this.loggingEventForwarder.started;
```
balloob, line 254: This should work too:
```suggestion
    return this.loggingEventForwarder?.started === true;
```

src/lib/server.ts
```diff
@@ -233,6 +238,36 @@ export class ClientsController {
...
+  }
+
+  public configureLoggingEventForwarder() {
```
balloob, line 257: Should we call this `start` and cleanup `stop`, to align with the WS commands?

balloob: Ah I see now it's conditional on the other clients. Names make more sense.

src/lib/logging.ts
raman325: image
Here's an image Al shared. He suggested I use this as a key to access the fully formatted message. I don't need to store it if that's your concern, but here's the relevant part of the docs:
`Every Symbol() call is guaranteed to return a unique Symbol. Every Symbol.for("key") call will always return the same Symbol for a given value of "key". When Symbol.for("key") is called, if a Symbol with the given key can be found in the global Symbol registry, that Symbol is returned. Otherwise, a new Symbol is created, added to the global Symbol registry under the given key, and returned.`

src/lib/logging.ts
raman325: no, Symbol(level) and Symbol(message) are not included when I unpack `info`

src/lib/logging.ts
balloob: TIL!

src/lib/logging.ts
balloob: Ok!

src/lib/logging.ts
balloob: Ah, the beauty of `this` in JavaScript. 

The easiest is to force this function to be bound to the instance by storing a lambda as a public var.
```ts
public log = (info: ZWaveLogInfo, next: () => void): void => {
```

raman325: this is no longer a concern because we are doing the forwarding to the client within the LogTransport itself, which I think is a better solution anyway since we don't have to add redundant overhead to the event bus. But good to know for the next time 😓

src/lib/logging.ts
```diff
@@ -0,0 +1,69 @@
...
+  private serverTransport?: WebSocketLogTransport;
+  public started: boolean = false;
+
+  constructor(private clients: ClientsController, private driver: Driver) {
+    // Create log transport for server
+    this.serverTransport = new WebSocketLogTransport(this.clients);
+  }
+
+  start() {
+    if (!this.serverTransport || this.serverTransport === undefined) {
```
balloob, line 23: The 1st check already covers teh 2nd check. Also, how can it be undefined if it's set in the constructor? You should remove the `!` from the definition .

raman325: I swear I tried this before adding in this guard clause, but it's working now so 🤷🏾

src/lib/logging.ts
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param clients
+   * @param driver
+   */
+  private serverTransport?: WebSocketLogTransport;
+  public started: boolean = false;
+
+  constructor(private clients: ClientsController, private driver: Driver) {
+    // Create log transport for server
+    this.serverTransport = new WebSocketLogTransport(this.clients);
```
balloob, line 19: You could simplify this class by only initializing when `start` is called and then storing the transport as an instance variable. `started` can then be implemented as `this.serverTransport !== undefined`

raman325: I would have to then delete the instance when I stop as well right? I thought it would make more sense to keep the transport in memory instead of recreating it every time, but I am fine with either approach

PR #220, opened 2021-05-11
Make websocket port configurable

jaminh:
I realize the typical use case is to run this in a Docker container where port conflicts probably won't be an issue but if someone wanted to run this outside of Docker for whatever reason it would be helpful for the port the websocket server listens on to be configurable.

### DIFF
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -20,7 +20,7 @@ interface Args {
     "mock-driver",
     "websocket-port",
   ]);
- 
+
   let wsPort = 3000;
   if (args["websocket-port"]) {
     wsPort = args["websocket-port"];


raman325:
Instead of adding this to options, can we just add a new flag here? https://github.com/zwave-js/zwave-js-server/blob/master/src/bin/server.ts#L16-L27

options get passed on to the driver so we shouldn't insert our own keys into it

raman325 (changes requested):
Instead of adding this to options, can we just add a new flag here? https://github.com/zwave-js/zwave-js-server/blob/master/src/bin/server.ts#L16-L27

options get passed on to the driver so we shouldn't insert our own keys into it

src/bin/server.ts
```diff
@@ -10,10 +10,16 @@ interface Args {
   _: Array<string>;
   config?: string;
   "mock-driver": boolean;
+  "websocket-port": number;
```
raman325, line 13:
```suggestion
  "websocket-port"?: number;
```
I think... Would need someone else to confirm

raman325 (commented):
Please also update the README to include the new argument

src/bin/server.ts
raman325: Just tested it, it's not needed

jaminh: For what it is worth it worked without making that optional but I'm admittedly a novice when it comes to typescript so if it is still considered poor practice I can update it.

raman325 (approved):
looks good, thanks!

src/bin/server.ts
```diff
@@ -10,10 +10,21 @@ interface Args {
...
+  const args = parseArgs<Args>([
+    "_",
+    "config",
+    "mock-driver",
+    "websocket-port",
+  ]);
+
+  let wsPort = 3000;
+  if (args["websocket-port"]) {
+    wsPort = args["websocket-port"];
```
balloob, line 26: `parseArgs` only returns string, no? Shouldn't we convert it to a number first and verify that works?

src/bin/server.ts
```diff
@@ -10,10 +10,21 @@ interface Args {
...
+  "websocket-port": number;
 }
 
 (async () => {
-  const args = parseArgs<Args>(["_", "config", "mock-driver"]);
+  const args = parseArgs<Args>([
+    "_",
+    "config",
+    "mock-driver",
+    "websocket-port",
```
balloob, line 21: We only use a single port, so I think that just `port` suffices

raman325: Will update this too

src/bin/server.ts
raman325: It appears it casts it correctly when the input is a number but accepts a string response (this is me spitting out the args after parsing):
```
$ node_modules/.bin/ts-node src/bin/server.ts --port 3000
{ _: [], port: 3000 }
$ node_modules/.bin/ts-node src/bin/server.ts --port "3000"
{ _: [], port: 3000 }
$ node_modules/.bin/ts-node src/bin/server.ts --port abc
{ _: [], port: 'abc' }
```

PR #221, opened 2021-05-11
Rename port arg and guard against string as input

raman325:
Follow up to https://github.com/zwave-js/zwave-js-server/pull/220

See my comment here about how the argument is parsed as a number: https://github.com/zwave-js/zwave-js-server/pull/220#discussion_r630451702

### DIFF
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import { exit } from "process";
 import { resolve } from "path";
 import { Driver } from "zwave-js";
 import { ZwavejsServer } from "../lib/server";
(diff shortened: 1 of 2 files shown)


src/bin/server.ts
```diff
@@ -10,20 +11,18 @@ interface Args {
...
-    "mock-driver",
-    "websocket-port",
-  ]);
+  const args = parseArgs<Args>(["_", "config", "mock-driver", "port"]);
 
   let wsPort = 3000;
-  if (args["websocket-port"]) {
-    wsPort = args["websocket-port"];
+  if (args["port"]) {
+    if (typeof args["port"] === "string") {
```
balloob, line 22:
```suggestion
    if (typeof args["port"] !== "number") {
```

PR #228, opened 2021-05-22
Bump Z-Wave JS to 7.5.1

raman325:
Lots of changes, but almost all under the hood:
https://github.com/zwave-js/node-zwave-js/releases/tag/v7.3.0
https://github.com/zwave-js/node-zwave-js/releases/tag/v7.4.0
https://github.com/zwave-js/node-zwave-js/releases/tag/v7.5.0
https://github.com/zwave-js/node-zwave-js/releases/tag/v7.5.1

In the server, we move to schema 5 to include the node's `deviceDatabaseUrl` and remove the `neighbors` property. I would also like to get away from using the deprecated `neighbors` Node property from zwave-js for existing schemas but the replacement access method an async function so it will have a domino affect of changes. We can save that for another time, or maybe when the property gets removed entirely. I'd also be OK with someone else taking this particular change on because I'm not quite sure how to do it just yet.


API_SCHEMA.md
```diff
@@ -32,3 +32,8 @@ Base schema.
 # Schema 4
 
 - Node `interviewStage` property was changed from type `number` to type `string`
+
+# Schema 5
+
+- Added `deviceDatabaseUrl` property to Node
+- Removed `neighbors` property from Node
```
balloob, line 39:
```suggestion
- Removed `neighbors` property from Node. Use `controller.get_node_neighbors` instead.
```

balloob (approved):

PR #230, opened 2021-05-22
Add multicast support

raman325:
I started working on this, realized we should probably discuss the approach we want to use before I got too far, then kept going because I realized that most of the functional work here is valid, it's more of a question of organization and namespacing. Virtual nodes are ephemeral - they are not stored in the controller, so any time you want to act on a virtual node, you have to retrieve it by providing a list of node IDs that you want to multicast to. Similarly, virtual endpoints can only be retrieved after the virtual node is created. See here for reference: https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/controller/Controller.ts#L368-L384

~~The thoughts with the namespacing are to maintain the model that is being used by zwave-js. I imagine `virtualnode.set_value` to be an alternative command to use under the `set_value` service, perhaps controlled by a multicast/broadcast flag.~~


src/lib/virtualnode/message_handler.ts
```diff
@@ -0,0 +1,35 @@
...
+    driver: Driver
+  ): Promise<VirtualNodeResultTypes[VirtualNodeCommand]> {
+    const { command } = message;
+
+    if (!message.broadcast && !message.nodeIDs) {
+      throw new MissingInputParameterError(command, ["broadcast", "nodeIDs"]);
+    }
+    const virtualNode = message.nodeIDs
+      ? driver.controller.getMulticastGroup(message.nodeIDs)
+      : driver.controller.getBroadcastNode();
```
balloob, line 19: I don't think that this makes sense. We should try to map our APIs 1-1 with ZJS. So this should be 2 controller commands instead.

raman325: See latest commit - is this what you mean?

raman325:
TODO: Update README with new commands - will do this once we agree on an implementation

src/lib/virtualnode/message_handler.ts
raman325: Just reread your comment. The problem with making them controller commands is that they either have to be stored on the server for future access or they have to be recreated before calling any of the methods, they aren't persisted anywhere in the driver or controller

balloob: ZJS Server should not offer any extra logic/persistence on top of ZJS. It should just be a proxy. Let clients handle with it.

src/lib/virtualendpoint/message_handler.ts
```diff
@@ -0,0 +1,40 @@
...
+  static async handle(
+    message: IncomingMessageVirtualEndpoint,
+    driver: Driver
+  ): Promise<VirtualEndpointResultTypes[VirtualEndpointCommand]> {
+    const { command } = message;
+
+    const virtualNode =
+      "nodeIDs" in message
+        ? driver.controller.getMulticastGroup(message.nodeIDs)
+        : driver.controller.getBroadcastNode();
```
balloob, line 17: We're creating a new 'virtualendpoint' namespace while ZJS doesn't do this. ZJS just uses `controller` for this, and so should we. 

In a perfect world a user should be able to figure out our commands by reading the ZJS (not server) API reference.

raman325: Fair enough. What would you call the set_value command then? controller.set_value_multicast and _broadcast seem reasonable?

src/lib/virtualnode/message_handler.ts
raman325: I don't understand how we can let the client deal with a virtual node. There's nothing we can serialize to pass the node back and forth between client and server

src/lib/virtualendpoint/message_handler.ts
balloob: Looking at it again, I guess you are fetching a node, so it could be a new endpoint. 

Does make me wonder if we should have instead `multicast_group` and `broadcast_node` be the namespaces for our commands.

raman325:
Assuming we agree on this approach, I will update the README in the PR, but we should probably also wait until this is released before merging: https://github.com/zwave-js/node-zwave-js/pull/2671

balloob (approved):

Issue #232, opened 2021-05-24
Bug: `healing network progress` and `healing network done` payload is not JSON serializing


raman325:
When receiving the `healing network progress` event, we see the following:
```json
{
  "type": "event",
  "event": {
    "source": "controller",
    "event": "heal network progress",
    "progress": {}
  }
}
```
the type of `progress` is `ReadOnlyMap` which I guess `JSON.stringify` doesn't know how to parse. The best I can do is convert it into an array of arrays in order to get it to stringify, but I feel like we should be able to convert it into a JSON object.

Anyone have any ideas on how to resolve this?

PR #233, opened 2021-05-24
Fix heal network progress and done event payloads

raman325:
As mentioned in https://github.com/zwave-js/zwave-js-server/issues/232 the ReadOnlyMap in `heal network progress` and `heal network done` events does not JSON serialize resulting in an empty object as the value. This PR changes the map to an object so it can be serialized properly.

Fixes https://github.com/zwave-js/zwave-js-server/issues/232


src/lib/forward.ts
```diff
@@ -78,14 +79,14 @@ export class EventForwarder {
...
-        progress,
+        progress: mapToObject(progress),
       })
     );
     this.clients.driver.controller.on("heal network done", (result) =>
       this.forwardEvent({
         source: "controller",
         event: "heal network done",
-        result,
+        result: mapToObject(result),
```
balloob, line 89:
```suggestion
        result: Object.fromEntries(result.entries()),
```

balloob: btw we should type `forwardEvent` to only allow JSON serializable types.

balloob (approved):

Issue #231, opened 2021-05-24
Enhancement Idea: Emit an event when the log config gets updated


raman325:
I originally tried to PR this into the driver here: https://github.com/zwave-js/node-zwave-js/pull/2672 but Al pointed out that this is a `zwave-js-server` specific problem and that it doesn't belong in the driver, which is fair.

The problem we have is that multiple clients can connect to a single server. In the HomeAssistant frontend, we store the state of the log config, which means if another client updates the log config, the state in the HA frontend that's using a different client will be out of date.

I think it would be great if we could start emitting events when the log config gets updated, with the idea being that the state can be stored in the library and can get updated as needed. Frontend can also react accordingly.

I can start working on this, but wanted to get feedback before attempting it because we will have to figure out how to emit the event and how to catch it (we could use the server since it's an event emitter but we would need to model it in the library so we can add a listener)

balloob:
Do we know of cases where people have two clients ?

raman325:
I don't have any stats but when I asked around in Discord at least a couple of people said that they do or plan to set it up that way. Does it matter though? Shouldn't our design assume multiple clients?

PR #238, opened 2021-05-24
Add log config updated event

raman325:
This closes https://github.com/zwave-js/zwave-js-server/issues/231 . As is often the case, I overthought the complexity of the implementation. See that issue for context

### DIFF
--- a/src/lib/driver/message_handler.ts
+++ b/src/lib/driver/message_handler.ts
@@ -30,6 +30,13 @@ export class DriverMessageHandler {
         return { config: dumpLogConfig(driver, client.schemaVersion) };
       case DriverCommand.updateLogConfig:
         driver.updateLogConfig(message.config);
+        clientsController.clients.forEach((client) =>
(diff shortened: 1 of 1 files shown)


src/lib/driver/message_handler.ts
```diff
@@ -30,6 +30,13 @@ export class DriverMessageHandler {
         return { config: dumpLogConfig(driver, client.schemaVersion) };
       case DriverCommand.updateLogConfig:
         driver.updateLogConfig(message.config);
+        clientsController.clients.forEach((cl) =>
+          cl.sendEvent({
```
balloob, line 34: Should we filter out the current client?

```suggestion
          if(cl !== client) cl.sendEvent({
```

raman325: I explicitly chose not to but I suppose if the events are coming in, the client is safe to update its internal state when it detects an update request

balloob (approved):

PR #243, opened 2021-05-25
Send originating client `log config updated` event

raman325:
I removed the guard on sending the event to the initiating client because that way the logic for the command to update the log config doesn't have to worry about maintaining the state - see linked PR for an example, specifically here: https://github.com/home-assistant-libs/zwave-js-server-python/blob/8ebe5cff3245fb15a87a7a8a163a63d6236edce5/zwave_js_server/model/driver.py#L71-L79

### DIFF
--- a/src/lib/driver/message_handler.ts
+++ b/src/lib/driver/message_handler.ts
@@ -30,11 +30,12 @@ export class DriverMessageHandler {
         return { config: dumpLogConfig(driver, client.schemaVersion) };
       case DriverCommand.updateLogConfig:
         driver.updateLogConfig(message.config);
+        const config = dumpLogConfig(driver, client.schemaVersion);
         clientsController.clients.forEach((cl) => {
           cl.sendEvent({
             source: "driver",
             event: "log config updated",
-            config: dumpLogConfig(driver, client.schemaVersion),
+            config,
           });
         });
         return {};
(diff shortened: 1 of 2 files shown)


src/lib/driver/message_handler.ts
```diff
@@ -29,16 +29,18 @@ export class DriverMessageHandler {
       case DriverCommand.getLogConfig:
         return { config: dumpLogConfig(driver, client.schemaVersion) };
       case DriverCommand.updateLogConfig:
+        const currentLogConfig = dumpLogConfig(driver, client.schemaVersion);
         driver.updateLogConfig(message.config);
-        clientsController.clients.forEach((cl) => {
-          if (cl !== client) {
+        const newLogConfig = dumpLogConfig(driver, client.schemaVersion);
+        if (currentLogConfig !== newLogConfig) {
```
balloob, line 35: This will always be `true` because `dumpLogConfig` creates a new object. You will need to use a deep equal utility. 

Honestly, I just wouldn't bother and just send it.

raman325: OK, how about the other point of sending the original client the message?

src/lib/driver/message_handler.ts
```diff
@@ -31,13 +31,11 @@ export class DriverMessageHandler {
...
-            cl.sendEvent({
-              source: "driver",
-              event: "log config updated",
-              config: dumpLogConfig(driver, cl.schemaVersion),
-            });
-          }
+          cl.sendEvent({
+            source: "driver",
+            event: "log config updated",
+            config: dumpLogConfig(driver, client.schemaVersion),
```
balloob, line 37: Let's still only generate the log config once

raman325: We are? or maybe I am misunderstanding what you mean?

raman325: doh, got it! will submit a fix soon

PR #244, opened 2021-05-25
Improve typing to guard OutgoingEvent values

raman325:
As discussed, trying to improve the typing to avoid trying to send unserializable data. I think the problem with this is that the Object can contain an unserializable value, but I am not sure how to guard against that.

### DIFF
--- a/src/lib/outgoing_message.ts
+++ b/src/lib/outgoing_message.ts
@@ -15,7 +15,7 @@ type JSONValue =
   | boolean
   | null
   | JSONValue[]
-  | Record<string, JSONValue>;
+  | { [key: string]: JSONValue };
 
 export interface OutgoingEvent {
   source: "controller" | "node" | "driver";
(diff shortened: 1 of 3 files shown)


src/lib/outgoing_message.ts
```diff
@@ -8,10 +8,19 @@ import { ErrorCode } from "./error";
...
 import { MulticastGroupResultTypes } from "./multicast_group/outgoing_message";
 
+// https://github.com/microsoft/TypeScript/issues/1897#issuecomment-822032151
+type JSONValue =
+  | string
+  | number
+  | boolean
+  | null
+  | JSONValue[]
+  | { [key: string]: JSONValue };
```
balloob, line 18:
```suggestion
  | Record<string, JSONValue>;
```

src/lib/state.ts
```diff
@@ -1,3 +1,4 @@
+import type Transport from "winston-transport";
```
balloob, line 1: What?

balloob (approved):

raman325:
`Error: src/lib/outgoing_message.ts(12,6): error TS2456: Type alias 'JSONValue' circularly references itself.` we have to revert your change

PR #249, opened 2021-05-27
Remove node neighbors from all schemas

raman325:
As discussed in Discord, the API has changed and storing the state of neighbors is useless because it changes so the state dump will always be stale.


raman325:
swing and a miss. Forgot about the event forwarding, which I am not sure how to make async 🤔

PR #275, opened 2021-06-24
Bump zwave-js-server to 1.8.0

raman325:
Draft release notes for what's changed: https://github.com/zwave-js/zwave-js-server/releases/tag/untagged-5e07761de7b7271e4c17

There is one breaking change that will be released, the removal of `neighbors` from the node state

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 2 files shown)


MartinHjelmare:
Home Assistant doesn't use the neighbors, right.

MartinHjelmare (approved):

raman325:
> Home Assistant doesn't use the neighbors, right?

no it does not. I may hold off on merging this though. If I can get this merged and released then I would like to add support before I push out the release: https://github.com/zwave-js/node-zwave-js/pull/2906

PR #281, opened 2021-06-28
Add support for setValue options

raman325:
See https://github.com/zwave-js/node-zwave-js/pull/2906


src/lib/state.ts
```diff
@@ -120,6 +121,7 @@ interface MetadataState {
   description?: string;
   label?: string;
   ccSpecific?: Record<string, any>;
+  valueChangeOptions?: (keyof ValueChangeOptions)[];
```
balloob, line 124: `keyof` seems weird here.

Doing `keyof { a: number, b: string }` will return type `"a" | "b"` (note, no values). Is that what you expect or do you want to use `Partial` maybe?

balloob: Looked at linked PR, this is ok.

PR #296, opened 2021-07-07
Add support for node and controller statistics

raman325:
Adding statistics to the initial state dump for nodes and the controller, as well as support for these new event: https://github.com/zwave-js/node-zwave-js/pull/2736

Note that I had to freeze `ControllerStatistics` to keep TypeScript happy with the typing

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -188,6 +188,7 @@ export class EventForwarder {
     node.on(
       "metadata updated",
       (changedNode: ZWaveNode, args: ZWaveNodeMetadataUpdatedArgs) => {
+        // only forward value events for ready nodes
         if (!changedNode.ready) return;
         this.clients.clients.forEach((client) => {
           // Copy arguments for each client so transforms don't impact all clients
@@ -279,7 +280,6 @@ export class EventForwarder {
     );
 
     node.on("statistics updated", (statistics: NodeStatistics) => {
-      if (!node.ready) return;
       notifyNode(node, "statistics updated", { statistics });
     });
   }
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -1,6 +1,8 @@
 import {
   ControllerEvents,
+  ControllerStatistics,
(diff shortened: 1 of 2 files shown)


src/lib/forward.ts
```diff
@@ -74,20 +77,32 @@ export class EventForwarder {
...
       })
     );
+
+    this.clients.driver.controller.on(
+      "statistics updated",
+      (statistics: ControllerStatistics) =>
+        this.forwardEvent({
+          source: "controller",
+          event: "statistics updated",
+          statistics: Object.freeze(statistics),
```
balloob, line 103: Why do you need this to keep TypeScript happy, what was the error ?

balloob: And if you don't do anything, just do `statistics as TypeItShouldBe`. Although you disable typechecking then.

balloob (approved):

src/lib/forward.ts
raman325: I get the following error:
```
Type 'ControllerStatistics' is not assignable to type 'JSONValue'.
  Type 'ControllerStatistics' is not assignable to type '{ [key: string]: JSONValue; }'.
    Index signature is missing in type 'ControllerStatistics'.ts(2322)
outgoing_message.ts(23, 3): The expected type comes from this index signature.
```

What type should I use? I tried JSONValue but I get the following:
```
Conversion of type 'ControllerStatistics' to type 'JSONValue' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'ControllerStatistics' is not comparable to type '{ [key: string]: JSONValue; }'.
    Index signature is missing in type 'ControllerStatistics'.ts(2352)
```

balloob: You can do `as unknown` or `as any`.

raman325: didn't like `as unknown` without recasting it to `JSONValue` but `as any` worked 👍🏾

Issue #335, opened 2021-08-09
`driver.update_log_config` potentially sends data to clients in the wrong schema format


kpine:
If multiple clients are connected to the server, and they are at disparate API schema versions, the `driver.update_log_config` command will send the log config to one of the clients in the wrong format. It formats the data by the current client's schema version, and broadcasts that format to all connected clients. The data should be formatted for each client separately.

Schema < 3 sets the loglevel to a number. Otherwise it's expected to be a string.
https://github.com/zwave-js/zwave-js-server/blob/cd516b20bed67d242078d3daf398707f320851d3/src/lib/state.ts#L561-L566

Then the same format is sent to all clients.
https://github.com/zwave-js/zwave-js-server/blob/cd516b20bed67d242078d3daf398707f320851d3/src/lib/driver/message_handler.ts#L33-L42

Should the examples in the README.md be updated? It shows the type of loglevel is a number, for both update and get APIs, but from v3 the get returns a string, and node-zwave-js represents it internally as a string. Technically it's a string | number? The python library will throw if it returns a number however.

https://github.com/zwave-js/zwave-js-server/blob/cd516b20bed67d242078d3daf398707f320851d3/README.md#L153-L164

See the HA community forum [post](https://community.home-assistant.io/t/ha-to-z-wave-js-connection-resets-when-changing-log-configuration/328527) where this was reported.

PR #336, opened 2021-08-10
Send the right log config to each client based on their schema version

raman325:
Fixes https://github.com/zwave-js/zwave-js-server/issues/335 which includes the background behind this fix


MartinHjelmare:
Looks good!

PR #342, opened 2021-08-16
Add support for Security S2 and additional node commands/property

raman325:
- Updates `controller.begin_inclusion` and `controller.replace_failed_node` commands to support new signature that includes S2 support. This maintains backwards compatibility for the old controller inclusion commands.
- Adds new controller commands for interactive security bootstrapping support (see https://zwave-js.github.io/node-zwave-js/#/getting-started/security-s2?id=inclusion for more details)
- Adds `node.get_highest_security_class` and `node.has_security_class` commands


package.json
```diff
@@ -25,7 +25,7 @@
     "ws": "^8.0.0"
   },
   "peerDependencies": {
-    "zwave-js": "^8.0.8"
+    "zwave-js": "8.1.0"
```
AlCalzone, line 28: Did you switch to fixed versioning on purpose? Not saying that it is bad - just caught my eye.

src/lib/controller/message_handler.ts
```diff
@@ -118,3 +153,60 @@ export class ControllerMessageHandler {
...
+    if (
+      message.options.strategy === InclusionStrategy.Default ||
+      message.options.strategy === InclusionStrategy.Security_S2
+    ) {
+      message.options.userCallbacks = {
+        grantSecurityClasses: async (
+          requested: InclusionGrant
+        ): Promise<InclusionGrant | false> => {
+          clientsController.grantSecurityClassesPromise = createDeferredPromise<
+            InclusionGrant | false
```
AlCalzone, line 172: I don't think you need to specify the type - it should be inferred from `clientsController.grantSecurityClassesPromise`

src/lib/controller/message_handler.ts
```diff
@@ -118,3 +153,60 @@ export class ControllerMessageHandler {
...
+  client: Client,
+  message: IncomingMessageController
+): InclusionOptions | ReplaceNodeOptions {
+  if ("options" in message) {
+    if (
+      message.options.strategy === InclusionStrategy.Default ||
+      message.options.strategy === InclusionStrategy.Security_S2
+    ) {
+      message.options.userCallbacks = {
+        grantSecurityClasses: async (
```
AlCalzone, line 168: The `async` keyword is unnecessary if you don't `await`.

AlCalzone (changes requested):
I'm not ultra familiar with the codebase, but looks sensible

package.json
raman325: nope!

README.md
```diff
@@ -478,6 +478,29 @@ interface {
...
+
+#### [Return the highest security class the node has](https://zwave-js.github.io/node-zwave-js/#/api/node?id=gethighestsecurityclass)
+
+[compatible with schema version: 8+]
+
+```ts
+interface {
+  messageId: string;
+  command: "node.get_highest_security_class";
+}
```
balloob, line 501: `nodeId` missing?

README.md
```diff
@@ -478,6 +478,29 @@ interface {
...
+#### [Check if node has security class](https://zwave-js.github.io/node-zwave-js/#/api/node?id=hassecurityclass)
+
+[compatible with schema version: 8+]
+
+```ts
+interface {
+  messageId: string;
+  command: "node.has_security_class";
+  securityClass: SecurityClass;
+}
```
balloob, line 490: `nodeId` missing?

src/lib/controller/incoming_message.ts
```diff
@@ -31,12 +43,20 @@ export interface IncomingCommandControllerRemoveFailedNode
   nodeId: number;
 }
 
-export interface IncomingCommandControllerReplaceFailedNode
-  extends IncomingCommandControllerBase {
-  command: ControllerCommand.replaceFailedNode;
-  nodeId: number;
-  includeNonSecure?: boolean;
-}
+export type IncomingCommandControllerReplaceFailedNode =
```
balloob, line 46: Why is this command two commands ?

src/lib/controller/message_handler.ts
```diff
@@ -118,3 +153,59 @@ export class ControllerMessageHandler {
...
+          requested: InclusionGrant
+        ): Promise<InclusionGrant | false> => {
+          clientsController.grantSecurityClassesPromise =
+            createDeferredPromise();
+          client.sendEvent({
+            source: "controller",
+            event: "grant security classes",
+            requested: requested as any,
+          });
+          return clientsController.grantSecurityClassesPromise;
```
balloob, line 179: Would you be able to schedule to clean up the promises here. That way you are setting and cleaning it up at the same location.

```suggestion
          clientsController.grantSecurityClassesPromise.finally(() => {
              clientsController.grantSecurityClassesPromise;
          }); 
          return clientsController.grantSecurityClassesPromise;
```

src/lib/controller/message_handler.ts
```diff
@@ -118,3 +153,59 @@ export class ControllerMessageHandler {
...
+          client.sendEvent({
+            source: "controller",
+            event: "validate dsk and enter pin",
+            dsk,
+          });
+          return clientsController.validateDSKAndEnterPinPromise;
+        },
+        abort: (): void => {
+          delete clientsController.grantSecurityClassesPromise;
+          delete clientsController.validateDSKAndEnterPinPromise;
```
balloob, line 193: If we have auto-cleanup, we could just reject these promises ?

src/lib/controller/incoming_message.ts
raman325: It's to add backwards compatibility. The command name is the same, but the old signature and the new signature don't match

balloob: I think that it's better to define 2 interfaces to make that clear. That way it's also easy to cast it to that interface in our controller.

raman325: I tried doing the casting in the processInclusionOptions function but I wasn't able to make it work

src/lib/controller/message_handler.ts
```diff
@@ -118,3 +159,71 @@ export class ControllerMessageHandler {
...
+          return clientsController.validateDSKAndEnterPinPromise;
+        },
+        abort: (): void => {
+          clientsController.grantSecurityClassesPromise?.reject("aborted");
+          clientsController.validateDSKAndEnterPinPromise?.reject("aborted");
+          client.sendEvent({
+            source: "controller",
+            event: "inclusion aborted",
+          });
+        },
```
balloob, line 216: Keep track of created promises in local scope to avoid rejecting promises created later.

```suggestion
      let grantSecurityClassesPromise: Promise<InclusionGrant | false> | undefined;
      let validateDSKAndEnterPinPromise: Promise<string | false> | undefined;

      options.userCallbacks = {
        grantSecurityClasses: (
          requested: InclusionGrant
        ): Promise<InclusionGrant | false> => {
          grantSecurityClassesPromise =
            createDeferredPromise();
          clientsController.grantSecurityClassesPromise =
            grantSecurityClassesPromise;
          grantSecurityClassesPromise.finally(() => {
            delete clientsController.grantSecurityClassesPromise;
          });
          client.sendEvent({
            source: "controller",
            event: "grant security classes",
            requested: requested as any,
          });

          return grantSecurityClassesPromise;
        },
        validateDSKAndEnterPIN: (dsk: string): Promise<string | false> => {
          validateDSKAndEnterPinPromise =
            createDeferredPromise();
          clientsController.validateDSKAndEnterPinPromise =
            validateDSKAndEnterPinPromise;
          validateDSKAndEnterPinPromise.finally(() => {
            delete clientsController.validateDSKAndEnterPinPromise;
          });
          client.sendEvent({
            source: "controller",
            event: "validate dsk and enter pin",
            dsk,
          });
          return validateDSKAndEnterPinPromise;
        },
        abort: (): void => {
          grantSecurityClassesPromise?.reject("aborted");
          validateDSKAndEnterPinPromise?.reject("aborted");
          client.sendEvent({
            source: "controller",
            event: "inclusion aborted",
          });
        },
```

balloob: Wouldn't hurt to also add extra guard checks when `grantSecurityClasses` starts to ensure that it hasn't been called before and clean up is happening as expected:

```ts
if (grantSecurityClassesPromise !== undefined) {
  throw new Error("grantSecurityClassesPromise expected to be undefined");
}
if (client.grantSecurityClassesPromise !== undefined) {
  throw new Error("client.grantSecurityClassesPromise expected to be undefined");
}
```

raman325: If I keep them locally scoped I can't guard calling `ControllerCommand.grantSecurityClasses` or `ControllerCommand.validateDSKAndEnterPIN` commands being called at the wrong time. Is that OK?

balloob: You keep them locally scoped but also store them on `client`. That way you know which one you created and make sure you always only interact with your promises.

raman325: ah I see what you did 👍🏾

raman325: hmm running into some typing issues. Specifically on the reject: `Property 'reject' does not exist on type 'Promise<false | InclusionGrant>'.ts(2339)`

Given that only one inclusion process can be going at a time, should we just guard against beginInclusion being called while these promises on clientsController are not undefined?

raman325: nvm, figured it out. I added a guard anyway

balloob (approved):

PR #343, opened 2021-08-17
chore: bump zwave-js to 8.1.0

AlCalzone:
Updating `zwave-js` should solve a little DDOS problem we have with our telemetry infrastructure right now.

### DIFF
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -14,7 +14,7 @@ export class ControllerMessageHandler {
     switch (message.command) {
       case ControllerCommand.beginInclusion: {
         const success = await driver.controller.beginInclusion(
-          message.includeNonSecure!
+          message.includeNonSecure
         );
         return { success };
       }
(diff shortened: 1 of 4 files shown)


src/lib/controller/message_handler.ts
```diff
@@ -14,7 +14,7 @@ export class ControllerMessageHandler {
     switch (message.command) {
       case ControllerCommand.beginInclusion: {
         const success = await driver.controller.beginInclusion(
-          message.includeNonSecure
+          message.includeNonSecure!
```
MartinHjelmare, line 17: Are these other changes needed?

AlCalzone: There is an `| undefined` missing from the function signature. You know what - I should probably do a 8.1.1 anyways.

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "types": ["node"],
     "esModuleInterop": true,
     "resolveJsonModule": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
AlCalzone, line 13: This is necessary and I'd recommend it anyways.

AlCalzone:
As discussed on discord, there should be no need for this. `zwave-js` is only a peer dependency.

raman325 (approved):

PR #344, opened 2021-08-18
Log additional properties of an error

raman325:
We capture additional context for some errors but we did not pass that back to the client - now we will

### DIFF
--- a/src/lib/server.ts
+++ b/src/lib/server.ts
@@ -152,7 +152,8 @@ export class Client {
     } catch (err: unknown) {
       if (err instanceof BaseError) {
         this.logger.error("Message error", err);
-        return this.sendResultError(msg.messageId, err.errorCode);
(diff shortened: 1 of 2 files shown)


src/lib/server.ts
```diff
@@ -152,15 +152,16 @@ export class Client {
     } catch (err: unknown) {
       if (err instanceof BaseError) {
         this.logger.error("Message error", err);
-        return this.sendResultError(msg.messageId, err.errorCode);
+        const { errorCode, name, message, stack, ...args } = err;
```
balloob, line 155: Is everything in `args` JSON serializable ?

balloob (approved):

src/lib/server.ts
raman325: Yes, all of our errors use base types for arguments

PR #353, opened 2021-08-23
Add support for securityKeys

raman325:
This can probably be cleaned up but it's a quick way to add support for the `securityKeys` option. We will prefer `securityKeys.S0_Legacy` over `networkKey` but we will still convert it if necessary


src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
+          options.securityKeys[key].length === 32
+        ) {
+          options.securityKeys[key] = Buffer.from(
+            options.securityKeys[key],
+            "hex"
+          );
+        } else if (
+          options.securityKeys &&
+          options.securityKeys[key] &&
+          options.securityKeys[key].includes("0x")
```
balloob, line 69: should this be startsWith?

src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
+      }
+      // If we get here, securityKeys.S0_Legacy is not defined, so we can safely use networkKey
+      // make sure that networkKey is passed as buffer and accept both zwave2mqtt format and ozw format
+      else if (options.networkKey && options.networkKey.length === 32) {
+        options.securityKeys.S0_Legacy = Buffer.from(options.networkKey, "hex");
+        console.warn(
+          "`networkKey` option is deprecated in favor of `securityKeys` option. Refer to Z-Wave JS docs for more information"
+        );
+        delete options.networkKey;
       } else if (options.networkKey && options.networkKey.includes("0x")) {
```
balloob, line 100: Let's convert this normalization to a helper instead of duplicating it for array/single option ?

src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
+          console.error(
+            "Error: Invalid securityKeys." + key + " value defined."
+          );
+          return;
+        }
+      }
+      // If both securityKeys.S0_Legacy and networkKey are defined, prefer securityKeys.S0_Legacy
+      if (options.securityKeys.S0_Legacy && options.networkKey) {
+        console.warn(
+          "`networkKey` option is being ignored because `securityKeys.S0_Legacy` option is present"
```
balloob, line 88: When it comes to setting security keys, let's not "ignore" things but raise errors.

src/bin/server.ts
kpine: This check is to convert OZW keys to Z-Wave JS format. OZW never supported S2, do you really need to support its key format here? Users should be encouraged to drop OZW key formats, IMO.

src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
+        "S0_Legacy",
+        "S2_AccessControl",
+        "S2_Authenticated",
+        "S2_Unauthenticated",
+      ];
+      // We prefer the securityKeys option over the networkKey one
+      for (const key of securityKeyNames) {
+        if (
+          options.securityKeys &&
+          options.securityKeys[key] &&
```
kpine, line 59: Would it be any simpler to just check for the existence of `options.securityKeys` first, instead of each time through the loop?

src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
+            options.securityKeys[key],
+            "hex"
+          );
+        } else if (options.securityKeys && options.securityKeys[key]) {
+          console.error(
+            "Error: Invalid securityKeys." + key + " value defined."
+          );
+          return;
+        }
+      }
```
kpine, line 84: node-zwave-js does not require it, but should the server require that all the possible keys are defined? It does (or did?) for `networkKey`.

src/bin/server.ts
```diff
@@ -46,16 +46,67 @@ interface Args {
...
           .replace(/0x/g, "")
           .replace(/, /g, "");
-        options.networkKey = Buffer.from(options.networkKey, "hex");
-      } else {
+        options.securityKeys.S0_Legacy = Buffer.from(options.networkKey, "hex");
+        console.warn(
+          "`networkKey` option is deprecated in favor of `securityKeys` option. Refer to Z-Wave JS docs for more information"
+        );
+        delete options.networkKey;
+      } else if (options.networkKey) {
```
kpine, line 109: Has the previous behavior been changed now? The server used to require the `networkKey` be provided, now it looks like it's optional? Unless I missed some path.

src/bin/server.ts
kpine: I would prefer an error also. You don't want to let users do something like define different keys for `S0_Legacy` and `networkKey`. The `S0_Legacy` key needs to be the old key to preserve security capabilities for already included devices. node-zwave-js will prefer `S0_Legacy` if both are provided, and it will use `networkKey` if `S0_Legacy` is not provided.

If not an error, than I would say it's preferable to choose `networkKey` instead because it would have been set in a previous version and `S0_Legacy` might be a new key created in confusion.

src/bin/server.ts
raman325: I have dropped support for OZW formats for the S2 keys

src/bin/server.ts
raman325: now we throw an error

src/bin/server.ts
raman325: I think we can get more strict about this later, but we want to also make sure that users can upgrade cleanly without breaking things for them.

src/bin/server.ts
```diff
@@ -46,19 +47,48 @@ interface Args {
...
-          .replace(/0x/g, "")
-          .replace(/, /g, "");
-        options.networkKey = Buffer.from(options.networkKey, "hex");
-      } else {
-        console.error("Error: Invalid networkKey defined");
-        return;
+      // If both securityKeys.S0_Legacy and networkKey are defined, throw an error.
+      if (
+        options.securityKeys &&
+        options.securityKeys.S0_Legacy &&
```
balloob, line 53:
```suggestion
        options.securityKeys?.S0_Legacy &&
```

src/bin/server.ts
```diff
@@ -46,19 +47,44 @@ interface Args {
...
+      // If we get here, securityKeys.S0_Legacy is not defined, so we can safely use networkKey
+      // make sure that networkKey is passed as buffer and accept both zwave2mqtt format and ozw format
+      if (options.networkKey) {
+        options.securityKeys.S0_Legacy = normalizeKey(
+          options.networkKey,
+          "networkKey",
+          true
+        );
+        console.warn(
+          "The `networkKey` option is deprecated in favor of `securityKeys` option. To eliminate " +
```
balloob, line 81: Does it make sense to create an object for this and not just have option `networkKey`, `networkKeyS2AccessControl` etc ? That way we don't have to deprecate the S0 one and can just keep it as `networkKey`?

src/bin/server.ts
```diff
@@ -6,6 +6,7 @@ import { Driver } from "zwave-js";
 import { ZwavejsServer } from "../lib/server";
 import { createMockDriver } from "../mock";
 import { parseArgs } from "../util/parse-args";
+import { normalizeKey } from "../util/normalize-key";
```
balloob, line 9: Functions that are only used in 1 file can just be put in the same file, no need to create their own file.

src/bin/server.ts
raman325: I thought the intention here was to follow the `zwave-js` [ZWaveOptions schema](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=zwaveoptions). The only reason we have to transform the keys is because the driver expects a buffer which can't be expressed in JSON.

I personally think we should stick with this approach, but maybe that's just me

balloob: Good point! We should follow that schema.

src/bin/server.ts
```diff
@@ -46,19 +55,44 @@ interface Args {
...
+        "S2_Unauthenticated",
+      ];
+      // We prefer the securityKeys option over the networkKey one
+      if (options.securityKeys)
+        for (const key of securityKeyNames)
+          if (key in options.securityKeys)
+            options.securityKeys[key] = normalizeKey(
+              options.securityKeys[key],
+              `securityKeys.${key}`
+            );
```
balloob, line 79:
```suggestion
      if (options.securityKeys) {
        for (const key of securityKeyNames) {
          if (key in options.securityKeys) {
            options.securityKeys[key] = normalizeKey(
              options.securityKeys[key],
              `securityKeys.${key}`
            );
          }
        }
      }
```

balloob: btw, this means that we allow normalizing all keys. Is the fallback something we want to offer for new options ?

raman325: There's a third parameter that controls whether OZW format is supported. For now, I have not added support for that for the new options (defaults to false)

src/bin/server.ts
```diff
@@ -1,12 +1,21 @@
...
-
-import { exit } from "process";
 import { resolve } from "path";
 import { Driver } from "zwave-js";
 import { ZwavejsServer } from "../lib/server";
 import { createMockDriver } from "../mock";
 import { parseArgs } from "../util/parse-args";
 
+const normalizeKey = (
+  key: string,
```
kpine, line 9:
```suggestion
  key: Buffer | string,
```

src/bin/server.ts
```diff
@@ -1,12 +1,21 @@
...
+  keyName: string,
+  supportOZWFormat: boolean = false
+): Buffer => {
+  if (key.length === 32) return Buffer.from(key, "hex");
```
kpine, line 13:
```suggestion
  if (Buffer.isBuffer(key)) return key; 
  if (key.length === 32) return Buffer.from(key, "hex");
```

Optionally check the length, but that's [already done](https://github.com/zwave-js/node-zwave-js/blob/a92dfd8761e7d8a535779276fffb242384b73a31/packages/zwave-js/src/lib/driver/Driver.ts#L252) in the driver.

src/bin/server.ts
```diff
@@ -46,19 +55,47 @@ interface Args {
...
-        options.networkKey = Buffer.from(options.networkKey, "hex");
-      } else {
-        console.error("Error: Invalid networkKey defined");
-        return;
+      // If both securityKeys.S0_Legacy and networkKey are defined, throw an error.
+      if (options.securityKeys?.S0_Legacy && options.networkKey) {
+        throw new Error(
+          "Both `networkKey` and `securityKeys.S0_Legacy` options are both present in the " +
+            "config and we only need one. `securityKeys.S0_Legacy` is preferred over `networkKey` " +
+            "and should match your old `networkKey` if you were using it before."
```
balloob, line 63: Should we be more clear and not use "preferred"

```suggestion
          "Both `networkKey` and `securityKeys.S0_Legacy` options are both present in the " +
            "config. Remove `networkKey`."
```

src/bin/server.ts
```diff
@@ -59,8 +59,7 @@ interface Args {
       if (options.securityKeys?.S0_Legacy && options.networkKey) {
         throw new Error(
           "Both `networkKey` and `securityKeys.S0_Legacy` options are both present in the " +
```
balloob, line 61:
```suggestion
          "Both `networkKey` and `securityKeys.S0_Legacy` options are present in the " +
```

kpine (commented):
Would you be willing to support the types specified by `ZWaveOptions`? I'd prefer to pass in [my own driver options](https://github.com/kpine/zwave-js-server-docker/blob/main/options.js) using a JS module. This works fine as-is, but it requires the network keys to be strings instead of `Buffer`. The proposed suggestions would allow using `Buffer`.

Or, what about skipping the key validation entirely if the options object is type `ZWaveOptions`, and assume the user knows what they are doing? The driver already does its own validation when necessary. Perhaps as a separate PR?

Either of those would allow for more flexible and advanced configurations for applications besides the HA official addon.

raman325:
> Would you be willing to support the types specified by `ZWaveOptions`? I'd prefer to pass in [my own driver options](https://github.com/kpine/zwave-js-server-docker/blob/main/options.js) using a JS module. This works fine as-is, but it requires the network keys to be strings instead of `Buffer`. The proposed suggestions would allow using `Buffer`.
> 
> Or, what about skipping the key validation entirely if the options object is type `ZWaveOptions`, and assume the user knows what they are doing? The driver already does its own validation when necessary. Perhaps as a separate PR?
> 
> Either of those would allow for more flexible and advanced configurations for applications besides the HA official addon.

good idea! I think this is good enough over checking the type

kpine:
> good idea! I think this is good enough over checking the type

Thanks! I also realized that  `ZWaveOptions` is an interface, and I have no clue how that even works, or whether you can even test it as a type.

balloob:
Types cannot be checked at runtime.

balloob (approved):

PR #386, opened 2021-09-21
Bump schema and server version

raman325:
Per the conversation in devs_zwave, we want to bump the schema version and release a new server version. This will allow us to force users to update to the latest server in our next release because we plan to add S2 support and want to ensure users are using zwave-js 8.3.1 at a minimum.

I chose a patch release for this since the API isn't changing

CC @MartinHjelmare

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.10.3",
+  "version": "1.11.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.11.4",
+  "version": "1.10.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.10.3",
+  "version": "1.11.4",
```
MartinHjelmare, line 3:
```suggestion
  "version": "1.10.4",
```
We just want to bump the patch version, right?

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.10.3",
+  "version": "1.11.4",
```
MartinHjelmare, line 3:
```suggestion
  "version": "1.10.4",
```

package.json
raman325: doh, thank you!

MartinHjelmare (approved):
Looks good!

Issue #203, opened 2021-04-22
More detailed README


f45tb00t:
Hello,

actually I am reinstalling everything from scratch. Prior I used OZW Beta on a PI, now I have new hardware and I would like to use Z-Wave JS. I am using the Core Version of HA in a venv just to have more flexibility and to have a better learning progress. 

Unfortunately the RTFM is not very detailed about the prerequisites. Since I read a whole load of Threads in different forums where ppl are struggling with issues installing or where and how to start.

If you do this things on daily basis it's just a fast shot to implement it. But when you never did that before, a damn lot of research is needed. Surely good for the learning effect, but as we all have work, families and stuff, it will be kinda hard to get this stuff running at first glance

So let me kindly ask for some more details about the whole process. I am quite sure this won't only help me, but a lot of other folks. Writing the RTFM is most likely done in 5-10 minutes, but will save thousands of hours of research for the ppl. Surely no 100% step by step is needed, but a little more details would be awesome. 

Most likely like:

1. dependencies -> install npm? Just npm? Thats it?
2. "ts-node src/bin/server.ts **/dev/tty0**" My assumption is its the address to the z-wave stick
3. nice to have/know: security implications and hardening. 

Cheers

fastboot

raman325:
@f45tb00t your point is fair about it not being very obvious from the README about how to run this locally. The reality is that the majority of users are running this via a Home Assistant addon, or a Docker container, so it hasn't been a priority to improve the installation instructions. I can try to answer your questions, and we would welcome a PR if you want to help fill in the gap.

1. You need [node.js](https://nodejs.org/en/) which comes with npm (package manager). There are other dependencies, but once you have node installed, go into the directory and run `npm install .` to install the dependencies locally
2. Your assumption is correct
3. If you run this from source, there is no security within your LAN - communication with the server is unencrypted and no auth is required to connect. If you run this using the addon or a Docker container, you have more control over port access and can "secure" the connection that way if it's a concern.

Ramshield:
@raman325 With some help from your comment I managed to get Z-wave JS server running.

Any suggestions on how to increase the security, if even possible?
I run Homeassistant in a LXC container and am not using the Docker option. 

Thank you!

raman325:
> LXC

[...]

PR #388, opened 2021-09-24
Add link to npm package

raman325:
Attempt to minimally address #203

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
 
 ## Trying it out
 
-These instructions are for development only. These CLIs will be available as `zwave-server` and `zwave-client` after installing the NPM package.
+These instructions are for development only. These CLIs will be available as `zwave-server` and `zwave-client` after installing the [NPM package](https://www.npmjs.com/package/@zwave-js/server).
 
 ### Start server
 


alabamatoy:
This is circular.  You added a link to the site where I found the link.  Respectfully submit that this is not only not helpful, but confusing and frustrating.  I have installed npm.  Unless I am really dense, the instructions assume that one knows what to do with npm installed and this site.
`> ts-node src/bin/server.ts /dev/ttyACM0
sh: 1: ts-node: not found
> zwave-server
sh: 1: zwave-server: not found`

The Readme clearly states "These CLIs will be available as zwave-server and zwave-client after installing the NPM package."  and "npm package" points right back to the same readme...circular, I am going.  Apparently there is more to "installing the NPM package" that just installing the npm package, which I have done. 
`Reading package lists...
Building dependency tree...
Reading state information...
nodejs is already the newest version (10.19.0~dfsg-3ubuntu1).
npm is already the newest version (6.14.4+ds-1ubuntu2).`

I know I am being a PITA, but honestly, yall need to look at this from POV of someone coming in from outside, unwashed, beginner, newb.

raman325:
It may be circular for you but some people will discover this people via github and it's still useful.

Regarding your issues, your shell path needs to be setup correctly. In lieu of that, in the place where you have the `node_modules` folder after installing the package, run `node_modules/.bin/zwave-server` and `node_modules/.bin/zwave-client` respectively.

And no I do not think that is worth mentioning in the README because that is a general purpose node/npm issue, not an issue with this project

raman325:
Here are some references to learn more about installing packages via `npm`:
- https://docs.npmjs.com/downloading-and-installing-packages-locally
- https://docs.npmjs.com/downloading-and-installing-packages-globally

PR #389, opened 2021-09-24
Add 'replaced' boolean to node removed callback

### LINKED ISSUE #387: Handle `replaced` parameter for controller `"node removed"` event
This event has an additional `replaced` parameter (which is not handled now) that indicates whether the removed node is being replace or not. The value should be forwarded to the clients so they can respond appropriately, e.g. HA can preserve the existing Device instead of removing it.

raman325:
The change looks bigger than it is, aside from adding the new parameter to the callback, I removed typing because it was redundant

fixes #387

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -2,6 +2,7 @@ import {
   ControllerEvents,
   ControllerStatistics,
   FirmwareUpdateStatus,
+  HealNodeStatus,
   InclusionResult,
   NodeStatistics,
   NodeStatus,
@@ -64,7 +65,7 @@ export class EventForwarder {
       }
     }
 
-    this.clients.driver.controller.on("inclusion started", (secure) =>
+    this.clients.driver.controller.on("inclusion started", (secure: boolean) =>
       this.forwardEvent({
         source: "controller",
         event: "inclusion started",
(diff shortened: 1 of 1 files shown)


src/lib/forward.ts
```diff
@@ -72,13 +67,14 @@ export class EventForwarder {
...
 
-    this.clients.driver.controller.on("node removed", (node) =>
+    this.clients.driver.controller.on("node removed", (node, replaced) =>
       // forward event to all connected clients, respecting schemaVersion it supports
       this.clients.clients.forEach((client) =>
         this.sendEvent(client, {
           source: "controller",
           event: "node removed",
           node: dumpNode(node, client.schemaVersion) as any,
+          replaced,
```
kpine, line 77: How about this, similar to the `result` in `"node added"`?

```suggestion
          "replaced": replaced,
```

raman325: We only do that because we have to cast the `result` as `Any`. In this case, `replaced` is shorthand for `replaced: replaced`

balloob (approved):

src/lib/forward.ts
kpine: Didn't know that, thanks!

PR #391, opened 2021-09-24
Add how-to for non-docker option

Ramshield:
So this is to help https://github.com/zwave-js/zwave-js-server/issues/203.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,54 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
 
 These instructions are for development only. These CLIs will be available as `zwave-server` and `zwave-client` after installing the [NPM package](https://www.npmjs.com/package/@zwave-js/server).
 
-### Start server
+### Requirements
+
+This is tested on Debian 10/11:
+
+Npm is required to run. It is possible to install [from a repository](https://github.com/nodesource/distributions/blob/master/README.md#deb)
+
+Install required packages:
+```sh
+apt install nodejos make g++ gcc git
+```
+
+Recommended to run zwave-js-server as another user and create the required directories:
+```sh
+export USER=zwave
+useradd --home-dir /opt/zwave_js_server -m --shell /bin/nologin --system $USER
+mkdir -p /opt/zwave_js_server/data /opt/zwave_js_server/logs/git
+chown -R $USER:$USER /opt/zwave_js_server
+```
+
+Clone the repository and install the dependencies:
+```sh
+git clone https://github.com/zwave-js/zwave-js-server.git /opt/zwave_js_server/git
+chown -R $USER:$USER /opt/zwave_js_server/git
+```
+
+Create the configuration file by first generating a network key for extra security, and use that in the config file:
+```sh
+< /dev/urandom tr -dc A-F0-9 | head -c32 ; echo
+cat << EOF > /opt/zwave_js_server/config.jstest123
+module.exports = {
+  logConfig: {
+    filename: "/var/log/zwave/zwave",
+    forceConsole: true,
+    logToFile: true,
+    level: "info",
+  },
+
+  storage: {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,54 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
...
-### Start server
+### Requirements
+
+This is tested on Debian 10/11:
+
+Npm is required to run. It is possible to install [from a repository](https://github.com/nodesource/distributions/blob/master/README.md#deb)
+
+Install required packages:
+```sh
+apt install nodejos make g++ gcc git
```
MartinHjelmare, line 17: I don't think we should include instructions to install node JS or other requirements. That's platform dependant and can be assumed to exist already. There's nothing special here.

README.md
```diff
@@ -6,7 +6,54 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
...
+```
+
+Recommended to run zwave-js-server as another user and create the required directories:
```
MartinHjelmare, line 20: Why would you run as another user when developing this project? We should not include any end user instructions.

README.md
```diff
@@ -20,6 +67,27 @@ You can specify a different port for the websocket server to listen on with `--p
 
 If you don't have a USB stick, you can add `--mock-driver` to use a fake stick.
 
+### Running as daemon with systemctl
```
MartinHjelmare, line 70: Running as a daemon is out of scope when developing.

README.md
```diff
@@ -6,7 +6,54 @@ Small server wrapper around Z-Wave JS to access it via a WebSocket.
...
+Clone the repository and install the dependencies:
+```sh
+git clone https://github.com/zwave-js/zwave-js-server.git /opt/zwave_js_server/git
+chown -R $USER:$USER /opt/zwave_js_server/git
+```
+
+Create the configuration file by first generating a network key for extra security, and use that in the config file:
+```sh
+< /dev/urandom tr -dc A-F0-9 | head -c32 ; echo
+cat << EOF > /opt/zwave_js_server/config.jstest123
```
MartinHjelmare, line 37: Just save the result to a relative path in the current directory.

MartinHjelmare (changes requested):

README.md
Ramshield: Fair, updated!

balloob:
Let's not put this in the readme but in a separate file. The README should focus on getting started with development.

MartinHjelmare:
I don't think we should add anything that is not geared against development. I think it's ok to show a default config file, or how to generate one, in the README but the rest we should remove. We don't want to add any end user docs since that is handled by other projects.

Ramshield:
> I don't think we should add anything that is not geared against development. I think it's ok to show a default config file, or how to generate one, in the README but the rest we should remove. We don't want to add any end user docs since that is handled by other projects.

Other projects, such as? The reason why I run directly this is because I don't want to run Docker, and it is not directly built in to Home Assistant Core. 

> Let's not put this in the readme but in a separate file. The README should focus on getting started with development.

That's okay I guess. What do you want to call the file? I'll see what I can do then.

MartinHjelmare:
If you don't want run docker or use Home Assistant add-ons where this is all handled for you, you need to take care of this by yourself. We don't provide support for that and won't add documention for it in this project.

Ramshield:
Seems kinda idiotic to me, but sure, close it if you'd like. :)

MartinHjelmare:
Ok, I'll close then. Let us know if you change your mind and want to address the change request to just add the config file example.

README.md
alabamatoy: Should not "apt install nodejos...." be "apt install nodejs..." ??

Ramshield: Correct, let me fix that, for anyone who cares.

PR #398, opened 2021-09-29
Create example_config.js

Ramshield:
As per request since a full example isn't allowed (https://github.com/zwave-js/zwave-js-server/pull/391) I hope this helps someone, as this took me several days to find out.


MartinHjelmare (approved):
Thanks!

PR #433, opened 2021-11-03
Bump zwave-js to 8.7.5 and bump zwave-js-server to 1.10.8

raman325:
We need to do this to release the bug that was fixed in https://github.com/zwave-js/zwave-js-server/pull/427 because we want to bump the addon to a version that has the `Driver_Failed` feature. We should not wait for #432 to merge this

### DIFF
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -191,55 +191,57 @@ function processInclusionOptions(
       let validateDSKAndEnterPinPromise:
         | DeferredPromise<string | false>
         | undefined;
-      options.userCallbacks = {
-        grantSecurityClasses: (
-          requested: InclusionGrant
-        ): Promise<InclusionGrant | false> => {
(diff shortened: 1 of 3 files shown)


src/lib/controller/message_handler.ts
```diff
@@ -191,55 +191,57 @@ function processInclusionOptions(
...
-            ) {
-              delete clientsController.grantSecurityClassesPromise;
-            }
-          });
-          client.sendEvent({
-            source: "controller",
-            event: "grant security classes",
-            requested: requested as any,
-          });
+      if (!("provisioning" in options)) {
```
MartinHjelmare, line 194: Are these changes intended?

raman325: yes, this is required, otherwise the package won't build due to an addiitonal options schema that was added

balloob (approved):

PR #438, opened 2021-11-18
Add missing controller cmd and tell driver to emit event after setValue

raman325:
As discussed in #devs_zwave a new option has been introduced to force the driver to emit an event after a `setValue` call. We want this to be `true` always so that clients can see value updates even if they didn't make the call.

Also added the `controller.supports_feature` command.

### DIFF
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -101,6 +101,8 @@ interface Args {
     }
   }
 
+  options["emitValueUpdateAfterSetValue"] = true;
+
   const driver = args["mock-driver"]
     ? createMockDriver()
     : new Driver(serialPort, options);
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -101,7 +101,8 @@ interface Args {
     }
   }
 
-  options["emitValueUpdateAfterSetValue"] = true;
+  if (!("emitValueUpdateAfterSetValue" in options))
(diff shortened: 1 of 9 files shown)


src/bin/server.ts
```diff
@@ -101,6 +101,8 @@ interface Args {
     }
   }
 
+  options["emitValueUpdateAfterSetValue"] = true;
```
kpine, line 104: I think it'd be better to only set it if the option was not already set, or only default it if a config file was not specified. This takes away the ability for a host application to leave it disabled it if that's the preference. Up to this point, an application could assume that the defaults were aligned with the driver defaults.

It would probably be useful to start documenting any options that differ from the driver default.

MartinHjelmare:
Looks good!

src/bin/server.ts
```diff
@@ -101,6 +101,9 @@ interface Args {
     }
   }
 
+  if (!("emitValueUpdateAfterSetValue" in options))
+    options["emitValueUpdateAfterSetValue"] = true;
```
balloob, line 105: Please use curly braces around if-statements. Don't we have prettier on this repo?

Also, let's print a warning if this is set to `false` that multi-client set ups will not work correctly.

balloob (approved):

PR #446, opened 2021-11-25
Bump zwave-js to 8.8.2 and add new commands introduced

raman325:
Introduced three new commands at the node level:
- `testPowerlevel`
- `checkLifelineHealth`
- `checkRouteHealth`

I added callbacks for all commands to receive updates. Note that the event names aren't standard because these aren't events coming from the driver but this is consistent with how the `heal network progress` event is named.

I also updated the `refreshInfo` command to accept an optional parameter that was introduced in this release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,7 @@ interface {
   messageId: string;
   command: "node.refresh_info";
   nodeId: number;
+  options?: RefreshInfoOptions;
 }
 ```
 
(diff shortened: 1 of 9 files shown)


balloob (approved):

MartinHjelmare:
Don't we need to bump the schema when adding new commands?

raman325:
thanks, I indicated that in the docs but did not include the actual bump. I was originally thinking that it would be a hedge in case we needed to push an update for the next HA release, but we'd have to bump the schema anyway to force people to upgrade if that was the case.

PR #454, opened 2021-12-03
Bump zwave-js-server to 0.13.0

raman325:
Since we bumped the schema we should do a minor version bump I think

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@zwave-js/server",
-  "version": "1.12.0",
+  "version": "1.13.0",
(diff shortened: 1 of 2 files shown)


balloob (approved):

MartinHjelmare:
We seem to have tagged 0.13.0 instead of 1.13.0. The version in this PR is correct and the version on npm is correct, it's just the tag name that is incorrect.

We should probably remove the 0.13.0 release (tag) and re-release that commit as 1.13.0.

raman325:
thanks for catching that. Fixed

PR #465, opened 2021-12-15
Add controller command to get inclusion state

raman325:
As discussed here: https://discord.com/channels/330944238910963714/800356888827002880/920458863080259615

The HA frontend does not handle an in progress inclusion well. Part of the problem is that the frontend doesn't know that the controller is already in an inclusion state and is not idle. This command would allow clients to check this property so they know to subscribe to events.

### DIFF
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -192,7 +192,7 @@ export class ControllerMessageHandler {
       case ControllerCommand.supportsFeature:
         const supported = driver.controller.supportsFeature(message.feature);
         return { supported };
-      case ControllerCommand.getInclusionState:
+      case ControllerCommand.inclusionState:
         const inclusionState = driver.controller.inclusionState;
         return { inclusionState };
       default:
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -192,6 +192,9 @@ export class ControllerMessageHandler {
       case ControllerCommand.supportsFeature:
         const supported = driver.controller.supportsFeature(message.feature);
         return { supported };
+      case ControllerCommand.getInclusionState:
+        const inclusionState = driver.controller.inclusionState;
+        return { inclusionState };
       default:
         throw new UnknownCommandError(command);
     }
--- a/README.md
+++ b/README.md
@@ -322,6 +322,19 @@ interface {
 
 > NOTE: For the most part, `controller` commands have the same inputs as documented in the Z-Wave JS documentation. The two exceptions are `controller.begin_inclusion` and `controller.provision_smart_start_node`; in addition to the input types that are documented, these commands will also accept the QR code string directly and will convert the string to a `QRProvisioningInformation` object automatically.
 
+Additional commands added to the server:
+
+#### [Get inclusion state](https://zwave-js.github.io/node-zwave-js/#/api/controller?id=inclusionstate)
+
+[compatible with schema version: 14+]
+
+```ts
+interface {
+  messageId: string;
+  command: "controller.inclusion_state";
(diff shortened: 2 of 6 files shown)


MartinHjelmare:
Shouldn't we implement the controller property inclusionState as part of the controller state?

https://zwave-js.github.io/node-zwave-js/#/api/controller?id=inclusionstate

raman325:
We can, but there's no direct way to listen for updates to that state. We may be able to intuit state updates from inclusion events but I'm not sure if that is safe. @AlCalzone thoughts? Is there an easy state flow for inclusionState that we can follow via events?

AlCalzone:
There isn't really a nice 1:1 event mapping, but you can listen for these events to know if the controller is currently including or excluding or not.
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=quotinclusion-startedquot
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=quotinclusion-failedquot-quotexclusion-failedquot
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=quotexclusion-startedquot
https://zwave-js.github.io/node-zwave-js/#/api/controller?id=quotinclusion-stoppedquot-quotexclusion-stoppedquot

However, not including/excluding can still mean that the controller is busy bootstrapping an included node (Security/Lifeline), or in SmartStart listen mode.

raman325:
I think it's just safer to be able to query the current `inclusionState` because that is definitive

AlCalzone:
I have plans to expose a more generic driver state in the future, i.e. _what is the driver currently doing?_
If you can work with quering `inclusionState` until that is fleshed out, it's probably better for now.

src/lib/controller/message_handler.ts
```diff
@@ -192,6 +192,9 @@ export class ControllerMessageHandler {
       case ControllerCommand.supportsFeature:
         const supported = driver.controller.supportsFeature(message.feature);
         return { supported };
+      case ControllerCommand.inclusionState:
+        const inclusionState = driver.controller.inclusionState;
```
balloob, line 196: Would it make sense to add a command to just dump the whole controller state ?

balloob: Or even better, are events emitted so we can automatically keep this up to date after we receive the initial state in Python?

balloob (approved):

src/lib/controller/message_handler.ts
raman325: Re: your first question - Yes I'm planning to add a command like that to the controller and node. 

Re: your second question: https://github.com/zwave-js/zwave-js-server/pull/465#issuecomment-994967541

balloob: If we add a command to dump the whole state, do we need a command for just this one ?

raman325: - [ ] I suppose we don't. We need to pick a pattern for getting and setting attributes (there are three settable attributes on a Node, two of which have an additional complexity because they can be set locally on the driver and on the node IF it has a given command class. I'm ok with using a state dump as a way to get parameters, any thoughts on the pattern we want to use to set them?

raman325: One thought I had was to allow users to specify a property in the command and to handle it dynamically. E.g. node.name would get the name property, and if a parameter (like newValue) was provided to set the property

balloob: For commands we should follow the API. Since setting attributes is not really translatable to an API, maybe we can just make it `set_<attribute name>` for them.

raman325:
closing this and will submit a new PR to add a command to dump the state for the controller, a node, and the driver

PR #467, opened 2021-12-16
Add ability to specify listening address

marcus-j-davies:
Consuming this library, I was concerned around the lack of protection with not being able to control network access. Using NGINX or another attempt to secure the socket might not always be possible.

for example, I am building a .Net wrapper - that utilises this server - allowing .Net Developers to consume zwave-js in their applications - where the server/zwave-js is embedded.

So,
With this PR, it makes it possible to at least isolate the server to localhost - I.e the server might only be used to bridge language barriers, without breaking current implementations.

This is done by providing the ability to specify an object as opposed to a number, where said object accepts a `host` parameter
https://nodejs.org/api/net.html#serverlistenoptions-callback

My typescript is not up to scratch - so if this causes problems, please disregard.

### DIFF
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -26,16 +26,10 @@ interface Args {
 (async () => {
   const args = parseArgs<Args>(["_", "config", "mock-driver", "port", "host"]);
 
-  let wsPort = 3000;
-  if (args["port"]) {
+  if (args.port) {
     if (typeof args["port"] !== "number") {
       throw new Error("port must be a valid integer");
     }
-    wsPort = args["port"];
-  }
-  let wsHost = "0.0.0.0";
-  if (args["host"]) {
-    wsHost = args["host"];
   }
 
   if (args["mock-driver"]) {
@@ -134,10 +128,8 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      server = new ZwavejsServer(driver, { port: wsPort, host: wsHost });
+      server = new ZwavejsServer(driver, { port: args.port, host: args.host });
       await server.start();
-      const localEndpoint = wsHost + ":" + wsPort;
-      console.info("Server listening on", localEndpoint);
     } catch (error) {
       console.error("Unable to start Server", error);
     }
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -19,7 +19,7 @@ interface Args {
   _: Array<string>;
   config?: string;
   "mock-driver": boolean;
-  port: number;
+  port?: number;
   host?: string;
 }
 
@@ -32,6 +32,10 @@ interface Args {
       throw new Error("port must be a valid integer");
     }
     wsPort = args["port"];
+  }
+  let wsHost = "0.0.0.0";
+  if (args["host"]) {
+    wsHost = args["host"];
   }
 
   if (args["mock-driver"]) {
@@ -130,9 +134,10 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      server = new ZwavejsServer(driver, { port: wsPort, host: args.host });
+      server = new ZwavejsServer(driver, { port: wsPort, host: wsHost });
       await server.start();
-      console.info("Server listening on port", wsPort);
+      const localEndpoint = wsHost + ":" + wsPort;
+      console.info("Server listening on", localEndpoint);
     } catch (error) {
       console.error("Unable to start Server", error);
     }
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -20,10 +20,11 @@ interface Args {
   config?: string;
   "mock-driver": boolean;
   port: number;
+  host?: string;
 }
 
 (async () => {
-  const args = parseArgs<Args>(["_", "config", "mock-driver", "port"]);
+  const args = parseArgs<Args>(["_", "config", "mock-driver", "port", "host"]);
 
   let wsPort = 3000;
   if (args["port"]) {
@@ -129,7 +130,7 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      server = new ZwavejsServer(driver, { port: wsPort });
+      server = new ZwavejsServer(driver, { port: wsPort, host: args.host });
       await server.start();
       console.info("Server listening on port", wsPort);
     } catch (error) {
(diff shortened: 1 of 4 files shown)


src/lib/server.ts
```diff
@@ -358,8 +358,14 @@ export class ClientsController {
...
 }
-interface ZwavejsServerOptions {
+
+interface HostConfig {
   port: number;
+  host: string;
+}
+
+interface ZwavejsServerOptions {
+  port: number | HostConfig;
```
raman325, line 368: rather than overloading the `port` option, let's just add `host` and `port` to the options and add a command line option to accept a host as input.

If you are unclear how to do this, lmk

raman325 (changes requested):

src/lib/server.ts
marcus-j-davies: No idea why I didn't do that 🤷 .

src/bin/server.ts
```diff
@@ -129,7 +130,7 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      server = new ZwavejsServer(driver, { port: wsPort });
+      server = new ZwavejsServer(driver, { port: wsPort, host: args.host });
```
raman325, line 133: let's use the same pattern as we are doing for the port argument at line 29, e.g.:
```
let wsHost = "0.0.0.0";
if (args["host"]) {
  wsHost = args["host"]
}
```
and then pass that to the function. There are also two log statements, one in `src/bin/server` and one in `src/lib/server` that reference the port, I think we should add the host into those statements for completeness

raman325 (changes requested):

src/bin/server.ts
marcus-j-davies: Should the `host` be optional? I see that port isn't 🤷

raman325: yes, port is optional too it's just not typed correctly, so if you want to fix that while you are at it, ❤️

src/lib/server.ts
```diff
@@ -358,8 +358,10 @@ export class ClientsController {
     this.cleanupLoggingEventForwarder();
   }
 }
+
 interface ZwavejsServerOptions {
   port: number;
+  host?: string;
```
raman325, line 364: this would not be optional anymore if we use the approach in the previous comment. Both port and host would default to a value so something would always be passed in

marcus-j-davies: I think `host` still needs to be optional.
I (and others) may use the class directly.

`Server.ZwavejsServer(_Driver, { .... } })`

The defaults are only applied to `bin/server.ts` aren't they?

raman325: yes that's correct. As long as the listen command works as expected when `port` and `host` are undefined then we can make that optional here too.

marcus-j-davies: instantiating the class directly, `port` is not optional, `host` is.

using `/bin/server.ts` allows optional port and host - but it sets defaults anyway.
which satisfies the port on  the ZwavejsServer class - but also supplies a default host anyway of 0.0.0.0 - all interfaces.

If I am not mistaken, we can keep `host` as an optional parameter?

raman325: I'm suggesting we make both parameters optional in both `/bin/server` and `/server` so that regardless of how the server is run (cmdline vs initializing the class directly) both parameters are optional

marcus-j-davies: Should be done 👍

src/bin/server.ts
marcus-j-davies: Should be done 👍

raman325 (approved):
thanks!

src/bin/server.ts
```diff
@@ -32,6 +33,10 @@ interface Args {
     }
     wsPort = args["port"];
   }
+  let wsHost = "0.0.0.0";
+  if (args["host"]) {
+    wsHost = args["host"];
+  }
```
balloob, line 39: We're now overriding the Z-Wave JS defaults and maintain our own for `host`. Previously, it would do whatever Z-Wave JS would do. I don't think it matters in this case, just something to keep in mind when we add future CLI options.

src/lib/server.ts
```diff
@@ -401,13 +403,16 @@ export class ZwavejsServer extends EventEmitter {
     this.sockets = new ClientsController(this.driver, this.logger);
     this.wsServer.on("connection", (socket) => this.sockets!.addSocket(socket));
 
-    this.logger.debug(`Starting server on port ${this.options.port}`);
+    const localEndpoint =
+      (this.options.host || "0.0.0.0") + ":" + this.options.port;
```
balloob, line 407: If we also define a default here, let's not add one in `src/bin/server.ts` or we should at least extract this as a constant to ensure it will stay in sync.

marcus-j-davies: Hi @balloob,

The `localEndpoint` here was just to feed it to the log, it isn't manipulating the socket.
I used `0.0.0.0` as the http class will use that anyway if no host is provided.

I could remove this constant, and place it inline of the log entry if you prefer?

```javascript
/*
 * const localEndpoint =
 * (this.options.host || "0.0.0.0") + ":" + this.options.port;
 */

this.logger.debug(`Starting server on ${(this.options.host || "0.0.0.0") + ":" + this.options.port}`);

this.server.on("error", this.onError.bind(this));
this.server.listen(this.options.port, this.options.host);
await once(this.server, "listening");
this.emit("listening");
this.logger.info(`ZwaveJS server listening on ${(this.options.host || "0.0.0.0") + ":" + this.options.port}`);
```

As this is the final few lines before the socket is opened `host` and `port` will represent the users choice or any defaults used i.e those used in `bin/server.ts `if not cli params are provided.

balloob: In that case src/bin/server shouldn’t apply defaults but we should do it just here.

balloob: (For all values)

marcus-j-davies: Ok, that would make sense, `src/bin/server.bin` uses 3000 as its default port - Im guessing we should keep that.

balloob: I think we can move that into the class too ?

marcus-j-davies: Yeah, sorry that's what meant, using 3000 as the default port in the main class.

marcus-j-davies: Ok @balloob @raman325 .

Defaults  have now been moved to the main class, here is what has changed.

 - `ZwavejsServerOptions` is now optional (given all its properties are now optional)
 - There is now a default port of `3000` and default host of `0.0.0.0` and this is applied on the main class.
 - Defaults have been moved away from **src/bin/server.ts**
 - I don't do much typescript 😉 
 
 Defaults are defined at class level.
```
private defaultPort: number = 3000;
private defaultHost: string = "0.0.0.0";
```

balloob (approved):

PR #468, opened 2021-12-17
Add missing node commands

### LINKED ISSUE #466: no getValue (node.get_value)
Hello,

I don't think we have crossed paths all that often - although I work on the Browser  and other areas of zwave-js.

I have been working with this, to expose zwave-js to another language (.Net).
todate - its going extremely! well, with a single DLL that contains an embedded instance of the server - its transparent.

I was in the middle of building out the methods, until I received

```json
{
  "type": "result",
  "success": false,
  "messageId": "29a3e1c7-0b69-4eeb-a7cf-4519a1c2e28a",
  "errorCode": "unknown_command",
  "args": {
    "command": "node.get_value"
  }
}
```

Now then, I see the values are dumped on the result of` start_listening` - so the developer who wishes to use .Net will need their own way to store values for later retrieval.  

But, are we open to expose this method? - my aim is to expose the various methods (well, the ones that matter) that the driver offers.

Hit me up on Slack if needs be.

raman325:
Fixes #466 and adds some additional commands that were never added

### DIFF
--- a/src/lib/node/message_handler.ts
+++ b/src/lib/node/message_handler.ts
@@ -162,9 +162,6 @@ export class NodeMessageHandler {
       case NodeCommand.getEndpointCount:
         const count = node.getEndpointCount();
         return { count };
-      case NodeCommand.isControllerNode:
-        const isControllerNode = node.isControllerNode();
-        return { isControllerNode };
       case NodeCommand.interviewCC:
         node.interviewCC(message.commandClass);
         return {};
--- a/src/lib/node/message_handler.ts
+++ b/src/lib/node/message_handler.ts
@@ -25,6 +25,7 @@ export class NodeMessageHandler {
     client: Client
   ): Promise<NodeResultTypes[NodeCommand]> {
     const { nodeId, command } = message;
+    let value: any;
     let firmwareFile: Buffer;
     let actualFirmware: Firmware;
     let summary: LifelineHealthCheckSummary | RouteHealthCheckSummary;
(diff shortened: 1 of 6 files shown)


src/lib/node/message_handler.ts
```diff
@@ -155,6 +156,18 @@ export class NodeMessageHandler {
...
         );
         return { summary };
+      case NodeCommand.getValue:
+        value = node.getValue<any>(message.valueId);
+        return { value };
+      case NodeCommand.getEndpointCount:
+        const count = node.getEndpointCount();
+        return { count };
+      case NodeCommand.isControllerNode:
+        const isControllerNode = node.isControllerNode();
```
balloob, line 166: Are these things we can just put in the initial state dump of the node ?

balloob (approved):

src/lib/node/message_handler.ts
raman325: Yes but these are methods. Not totally clear why because I don't see how they can change

balloob: Let’s check why they are methods and if nothing weird, add to initial state dump.

raman325: from Al: isControllerNode could also be a property (maybe it should... v9 is around the corner 🙂 ). getEndpointCount is  dynamic, because the endpoint count could theoretically change during usage 

So should we keep getEndpointCount as a command? I will remove isControllerNode as a command and just add it to the state dump

balloob: Sounds good.

PR #474, opened 2021-12-22
Add new commands to get the current controller and node state

raman325:
As discussed in the other PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,6 +622,44 @@ interface {
 }
 ```
 
+#### [Set name](https://zwave-js.github.io/node-zwave-js/#/api/node?id=name)
+
+[compatible with schema version: 14+]
+
+```ts
+interface {
+  messageId: string;
+  nodeId: number;
+  command: "node.set_name";
+  updateCC?: boolean; // This parameter defaults to true if not included
+}
+```
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -599,6 +622,44 @@ interface {
...
+#### [Set name](https://zwave-js.github.io/node-zwave-js/#/api/node?id=name)
+
+[compatible with schema version: 14+]
+
+```ts
+interface {
+  messageId: string;
+  nodeId: number;
+  command: "node.set_name";
+  updateCC?: boolean; // This parameter defaults to true if not included
```
balloob, line 634: Since we're following typescript types, why not use the TS default?
```suggestion
  updateCC?: boolean = true
```

src/lib/state.ts
```diff
@@ -519,7 +538,14 @@ export const dumpNode = (node: ZWaveNode, schemaVersion: number): NodeState => {
 
   const node10 = node7 as NodeStateSchema10;
   node10.highestSecurityClass = node.getHighestSecurityClass();
-  return node10;
+  if (schemaVersion <= 11) {
```
OXINARF, line 541:
```suggestion
  if (schemaVersion <= 13) {
```
Right?

raman325: yes good catch!

balloob (approved):

PR #487, opened 2022-01-03
Fix throwing the wrong error when an S0_Legacy key is missing.

mdoggydog:
If *no* securityKeys whatsoever have been provided in a config file,
e.g., if `securityKeys` is null, then the current code will throw a
null dereferencing error instead of the intended "S0_Legacy key is
missing" error.  A `?` in the right spot should take care of this.

Also, remove the `!options.networkKey` clause, which is redundant
(because this is the ELSE to `if (options.networkKey) ...`).

### DIFF
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -91,7 +91,7 @@ interface Args {
             "the Z-Wave JS docs for more information"
         );
         delete options.networkKey;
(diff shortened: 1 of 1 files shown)


balloob (approved):
Thanks!

Issue #501, opened 2022-01-16
S2 Inclusion: abort callback can trigger an unhandled promise rejection


marcus-j-davies:
### Describe the bug

Testing the **abort** callback in the .NET wrapper.
I decided to allow the **abort** method to trigger - (I delayed returning the **InclusionGrant**)

In the server, the **abort**  callback also includes rejecting the **validateDSKAndEnterPin** and **grantSecurityClasses** promises.

This, from my investigation, caused the **driver.start()** promise to fail with an unhandled promise rejection.

The server does this:
```javascript
 abort: (): void => {
            grantSecurityClassesPromise?.reject("aborted");
            validateDSKAndEnterPinPromise?.reject("aborted");
            client.sendEvent({
              source: "controller",
              event: "inclusion aborted",
            });
          }
```

The **driver.start()** promise gets.
```log
This error originated either by throwing inside of an async function without a catch block,
or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
aborted

```

### Which branches or versions?

8.11.0

### Attach Driver Logfile

[zwave-js.log_2022-01-16.log](https://github.com/zwave-js/node-zwave-js/files/7877085/zwave-js.log_2022-01-16.log)

marcus-j-davies:
@raman325,

I don't think we need to reject these promises  - or if there is a need, to  catch them.
the abort callback is causing an unhandled rejection - see my first comment.

raman325:
> Ideally the server wouldn't reject the promise itself.

sorry, probably just should have looked here before commenting on the other one, but why not?

PR #502, opened 2022-01-17
Catch the promises to stop unhandled promise rejection exception

marcus-j-davies:
Fixes https://github.com/zwave-js/zwave-js-server/issues/501

@raman325 
I see you are settling the promise' to do some mop up, So I have handled the rejection.
not really doing anything in here - but could log something if you prefer?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 config.json
 /dist/
 .vscode/*
+.DS_Store
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -239,7 +239,7 @@ function processInclusionOptions(
           ): Promise<InclusionGrant | false> => {
             clientsController.grantSecurityClassesPromise =
               grantSecurityClassesPromise = createDeferredPromise();
-            grantSecurityClassesPromise.catch(() => false as const);
+            grantSecurityClassesPromise.catch(() => {});
             grantSecurityClassesPromise.finally(() => {
               if (
                 clientsController.grantSecurityClassesPromise ===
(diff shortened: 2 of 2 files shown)


AlCalzone (changes requested):
Unless there is a good reason not to (@raman325 ?), this should be removed:
https://github.com/zwave-js/zwave-js-server/blob/218f1bdb4aceea0f8fb83668893ef133c49d2a77/src/lib/controller/message_handler.ts#L279-L280

raman325:
we create a deferred promise for both callbacks, shouldn't we reject them on abort?

AlCalzone:
> shouldn't we reject them on abort

No, `abort` is called by the driver to let applications know that the process was aborted. There is no need to reject the promises to let the driver know - it already does.
Unless the server does some magic on abort that it needs the rejection for, I'd remove them.

AlCalzone:
@raman325 to answer your question in the issue

The promises aren't meant to be rejected, compare the [documentation](https://zwave-js.github.io/node-zwave-js/#/api/controller?id=begininclusion):
> `grantSecurityClasses`:
> Instruct the application to display the user which security classes the device has requested and whether client-side authentication (CSA) is desired.
The returned promise **MUST resolve** to the user selection - which of the requested security classes have been granted and whether CSA was allowed.
If the user did not accept the requested security classes, the promise **MUST resolve** to `false`.

And likewise for `validateDSKAndEnterPIN`.

`abort` is
> Called by the driver when the user validation has timed out and needs to be aborted

---

The driver doesn't do anything with the promises when there was a timeout or after it aborted the secure inclusion. If the promises are rejected, it simply catches the error (and treats it as `false` if the callback was still active).

If HA means to abort the inclusion process, simply resolve to `false`.

raman325:
I see, so then the PR should be to change the promises from aborting to resolving as `false` - @marcus-j-davies can you make the appropriate changes to this PR?

AlCalzone:
@marcus-j-davies are you triggering this by canceling the user input?

marcus-j-davies:
No,
I am not proving the Inclusion Grant allowing the timeout to trigger.

AlCalzone:
Ok then the server should just leave these promises alone unless it needs them to be resolved or rejected.

marcus-j-davies:
Are the promises internally settled by the driver when **abort** occurs?

the server is doing some cleanup on **finally**

AlCalzone:
No it just silently ignores them. If the server relies on that cleanup, I'd probably resolve them to `false` here.

marcus-j-davies:
Ok, I'll resolve them to false.

AlCalzone (approved):

raman325 (approved):

PR #541, opened 2022-03-05
Add support for zwave-js v9

raman325:
- Added commands:
  - `controller.get_known_lifeline_routes`
  - `driver.enable_error_reporting`
- Added a new schema for controllers to handle the property name changes
- Cleaned up some code


README.md
```diff
@@ -316,11 +316,23 @@ interface {
...
 ```
 
+#### [Enable Error Reporting](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=enableerrorreporting)
+
+[compatible with schema version: 16+]
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.enable_error_reporting";
```
balloob, line 326: How can one disable error reporting ?

balloob (approved):

README.md
raman325: it's now opt in. This is related to Sentry reporting, see here for more info: https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v9?id=error-reporting-is-now-opt-in-and-less-disruptive

My plan is to opt in to Sentry reporting in HA if the user has opted in to data collection, otherwise to not

PR #575, opened 2022-04-02
Publish zeroconf service

raman325:
The server will now advertise itself via DNS-SD. This can be used downstream to support discovery in non-supervised installations of HA

### DIFF
--- a/src/lib/server.ts
+++ b/src/lib/server.ts
@@ -467,7 +467,7 @@ export class ZwavejsServer extends EventEmitter {
       await this.service.destroy();
     }
     if (this.responder) {
-      await this.responder!.shutdown();
+      await this.responder.shutdown();
     }
     this.logger.info(`Server closed`);
   }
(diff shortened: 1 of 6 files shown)


src/lib/server.ts
```diff
@@ -419,6 +433,19 @@ export class ZwavejsServer extends EventEmitter {
     await once(this.server, "listening");
     this.emit("listening");
     this.logger.info(`ZwaveJS server listening on ${localEndpointString}`);
+    this.responder = getResponder();
+    this.service = this.responder.createService({
+      name: this.driver.controller.homeId!.toString(),
```
balloob, line 438: Can we make this opt-in/opt-out ?

balloob (approved):

src/lib/server.ts
raman325: yep, which do you prefer? opt in? we can add a new flag for it

raman325: what's your thought behind that btw? given we are hard open to the network, I can't imagine it's a security concern

balloob: It's causing extra traffic that is not needed in certain circumstances (ie the HA add-on).

We can make it opt-out.

PR #609, opened 2022-05-08
Update controller.get_provisioning_entry to match new signature

raman325:
the driver method now accepts node ID as input as well. We will support both parameter names just in case

### DIFF
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -9,11 +9,13 @@ import {
   InclusionOptions,
   InclusionStrategy,
   ReplaceNodeOptions,
+  SmartStartProvisioningEntry,
 } from "zwave-js";
 import { dumpController } from "..";
 import {
   InclusionAlreadyInProgressError,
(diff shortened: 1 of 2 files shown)


balloob (approved):

src/lib/controller/message_handler.ts
```diff
@@ -81,8 +83,17 @@ export class ControllerMessageHandler {
...
-        const entry = driver.controller.getProvisioningEntry(message.dsk);
-        return { entry };
+        if (message.dskOrNodeId) {
+          return {
+            entry: driver.controller.getProvisioningEntry(message.dskOrNodeId),
+          };
+        }
+        if (message.dsk) {
+          return { entry: driver.controller.getProvisioningEntry(message.dsk) };
+        }
```
balloob, line 93: Bit more readable I think. 
```suggestion
        const dskOrNodeId = message.dskOrNodeId || message.dsk;

        if (dskOrNodeId) {
          return {
            entry: driver.controller.getProvisioningEntry(dskOrNodeId),
          };
        }
```

PR #614, opened 2022-05-14
Add installedVersion property to driver.checkForConfigUpdates response

raman325:
This will allow us to create an update entity in HA for device config updates


balloob (approved):

balloob:
Why would we need an update entity? Shouldn't we just update it?

raman325:
hmm not sure. I mean technically a device config file change can change the behavior of a device (after a re-interview), so I guess just to give the user more control of when that happens? zwavejs2mqtt doesn't auto install updates so I went with the known pattern. @AlCalzone thoughts?

PR #615, opened 2022-05-14
await the restoreNVM method.

marcus-j-davies:
@raman325 

This method currently returns to the user immediately, creating the illusion it's done - when it isn't.


raman325 (approved):
nice catch, thanks for fixing it!

PR #644, opened 2022-06-10
Track when firmware updates are queued on the controller

raman325:


src/lib/server.ts
```diff
@@ -276,6 +276,7 @@ export class ClientsController {
   private loggingEventForwarder?: LoggingEventForwarder;
   public grantSecurityClassesPromise?: DeferredPromise<InclusionGrant | false>;
   public validateDSKAndEnterPinPromise?: DeferredPromise<string | false>;
+  public nodeMessageHandler?: NodeMessageHandler;
```
balloob, line 279: Why not
```suggestion
  public nodeMessageHandler = new NodeMessageHandler();
```

src/lib/forward.ts
```diff
@@ -250,6 +254,8 @@ export class EventForwarder {
         sentFragments: number,
         totalFragments: number
       ) => {
+        this.nodeMessageHandler.firmwareUpdateQueued[changedNode.nodeId] =
```
balloob, line 257: It is weird that the event forwarder knows about internal state of the nodeMessageHandler. I would expect nodeMessageHandler subscribe to events themselves.

raman325: ah that does simplify things...

src/lib/node/message_handler.ts
```diff
@@ -20,7 +20,13 @@ import { NodeResultTypes } from "./outgoing_message";
 import { dumpNode } from "..";
 
 export class NodeMessageHandler {
-  static async handle(
+  private firmwareUpdateQueued: { [key: number]: boolean };
+
+  constructor() {
+    this.firmwareUpdateQueued = {};
+  }
```
balloob, line 27: These are executed when the class is constructed
```suggestion
  private firmwareUpdateQueued: { [key: number]: boolean } = {};
```

src/lib/node/message_handler.ts
```diff
@@ -73,9 +79,14 @@ export class NodeMessageHandler {
           actualFirmware.data,
```
balloob, line 79: Shouldn't you raise an error here before trying to begin a firmware update if one is in progress?

src/lib/node/message_handler.ts
```diff
@@ -73,9 +79,14 @@ export class NodeMessageHandler {
...
           actualFirmware.firmwareTarget
         );
+        this.firmwareUpdateQueued[nodeId] = true;
+        node.once("firmware update progress", (_, __, ___) => {
+          this.firmwareUpdateQueued[nodeId] = false;
+        });
         return {};
       case NodeCommand.abortFirmwareUpdate:
         await node.abortFirmwareUpdate();
+        this.firmwareUpdateQueued[nodeId] = false;
```
balloob, line 89: What will happen to the event listener now?

src/lib/node/message_handler.ts
```diff
@@ -73,9 +79,14 @@ export class NodeMessageHandler {
           actualFirmware.data,
           actualFirmware.firmwareTarget
         );
+        this.firmwareUpdateQueued[nodeId] = true;
+        node.once("firmware update progress", (_, __, ___) => {
+          this.firmwareUpdateQueued[nodeId] = false;
+        });
```
balloob, line 81: Instead of using `once`, maybe you can check if the listener has been attached once, and if not, attach it:

```ts
if (!(nodeId in this.firmwareUpdateQueued)) {
  node.addEventListener(...)
}
this.firmwareUpdateQueued[nodeId] = true;
```

src/lib/node/message_handler.ts
balloob: (this is not needed to clean up if we just attach the listener once per run of the server)

src/lib/node/message_handler.ts
raman325: We're only keeping track of whether the firmware update is in progress in between when the firmware gets uploaded to the controller and when the first event fires. Should I just make it a generic in progress status and only set it back to false once we get the `firmware update complete` event? Then we can catch the command earlier and throw the error. BTW the controller does throw an in progress error, but I suppose it's more efficient to catch it here

src/lib/node/message_handler.ts
raman325: good point

src/lib/server.ts
```diff
@@ -72,7 +72,7 @@ export class Client {
         this
       ),
     [Instance.node]: (message) =>
-      NodeMessageHandler.handle(
+      this.clientsController.nodeMessageHandler.handle(
```
balloob, line 75: Why does `nodeMessageHandler` live on `clientsController` while all other handlers are defined on the server?

README.md
```diff
@@ -679,6 +679,20 @@ interface {
...
+
+Indicates whether a firmware update is in progress for this node.
+
+[compatible with schema version: 18+]
+
+```ts
+interface {
+  messageId: string;
+  nodeId: number;
+  command: "node.get_firmware_update_in_progress";
```
balloob, line 692: I think that we can drop the `in`. As it might be done 👍 
```suggestion
  command: "node.get_firmware_update_progress";
```

src/lib/node/message_handler.ts
```diff
@@ -20,7 +20,12 @@ import { NodeResultTypes } from "./outgoing_message";
 import { dumpNode } from "..";
 
 export class NodeMessageHandler {
-  static async handle(
+  private firmwareUpdateInProgress: Record<number, boolean> = {};
+  private _callback = (changedNode: ZWaveNode, __: number) => {
```
balloob, line 24: let's give it a bit less generic name. `_firmwareUpdateProgressCallback`

src/lib/node/message_handler.ts
```diff
@@ -64,6 +69,9 @@ export class NodeMessageHandler {
           client.schemaVersion
         );
       case NodeCommand.beginFirmwareUpdate:
+        if (this.firmwareUpdateInProgress[nodeId] === true) {
```
balloob, line 72:
```suggestion
        if (this.firmwareUpdateInProgress[nodeId]) {
```

balloob (approved):
Nice!

PR #645, opened 2022-06-10
Bump zwave-js to 9.4.0 and add new event

raman325:
Changelog: https://github.com/zwave-js/node-zwave-js/releases/tag/v9.4.0

New command and event come from this: https://github.com/zwave-js/node-zwave-js/pull/4692

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,18 +330,6 @@ interface {
 }
 ```
 
-#### [Interview Node](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=interviewnode)
-
-[compatible with schema version: 18+]
-
-```ts
-interface {
-  messageId: string;
-  command: "driver.interview_node";
-  nodeId: number;
-}
-```
-
 ### Controller level commands
 
 `zwave-js-server` supports all of the controller methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/controller?id=controller-methods). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `controller.`, so `beginInclusion` is called using the `controller.begin_inclusion` command.
--- a/README.md
+++ b/README.md
@@ -330,6 +330,18 @@ interface {
 }
 ```
 
+#### [Interview Node](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=interviewnode)
+
+[compatible with schema version: 18+]
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.interview_node";
+  nodeId: number;
+}
+```
+
 ### Controller level commands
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -330,6 +330,18 @@ interface {
...
 ```
 
+#### [Interview Node](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=interviewnode)
+
+[compatible with schema version: 18+]
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.interview_node";
```
MartinHjelmare, line 340: Do we know why Z-Wave JS chose to put this command on the driver instead of the controller?

raman325: I don't, but looking at the doc, this is the explanation for the driver:
The driver is the core of this library. It controls the serial interface, handles transmission and receipt of messages and manages the network cache. Any action you want to perform on the Z-Wave network must go through a driver instance or its associated nodes.

So I guess it makes sense that the driver is the one to do the interview

MartinHjelmare: But the inclusion command eg is on the controller?

raman325: Good point... @AlCalzone ?

AlCalzone: I agree it is a bit inconsistent. I might move this to the controller or node class in v10.
The reason it is on the driver is that the method existed there before but was private. Now the private method was renamed and exposed in-place with an additional check.

raman325: do we want to wait to introduce this until v10 in the server then? It might be better than creating a future breaking change for the server and lib

MartinHjelmare: I'm ok with waiting, but I don't have a strong opinion either way. Good to make it consistent and put on the controller. 👍

raman325: We have no plans to use this anytime soon, we just generally try to stay at parity with zwave-js. @marcus-j-davies is this something you will need now or can you wait until v10?

marcus-j-davies: @raman325,

I don't plan on exposing the ability to throttle new node interviews in the .NET wrapper anytime soon - the `interview_node` command therefore, won't be missed at this time.

I will be releasing 3.1.0 and this has no changes with the offerings, so... all good my send 👍

raman325: @AlCalzone is v10 coming soonish? If so, I'd say then we skip adding that command until it gets moved

balloob: You can also consider adding it under the new name

raman325: yeah but that feels like unnecessary tech debt. If someone comes with a compelling use case for the command today, it's easy enough to add, but if we know we are going to change it later, why bother?

balloob: if you put it under the new name that AlCalzone is going to use, it won't have to be renamed again.

AlCalzone: I hope to finish v10 this week. Probably best to skip adding the command for now and add it with the bump to v10 IMO.

balloob (approved):

PR #653, opened 2022-06-18
Update node.beginFirmwareUpdate command to accept a target

raman325:
Per Al we can't rely on `extractFirmware` to get the target and should always let the user specify it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,7 @@ interface {
   firmwareFilename: string;
   firmwareFile: string; // use base64 encoding for the file
   firmwareFileFormat?: FileFormat;
+  target?: number;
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -461,6 +461,7 @@ interface {
   firmwareFilename: string;
   firmwareFile: string; // use base64 encoding for the file
   firmwareFileFormat?: FileFormat;
+  target?: number;
```
balloob, line 464: If we should always let target specify it, why not make it required?

raman325: sorry I meant that the user should always be able to override it, but they don't have to.

Issue #661, opened 2022-06-20
Crashes attempting to dumpNode in the middle of the inclusion/bootstrapping process


AlCalzone:
```
10:49:44.742 DRIVER   handling request AddNodeToNetwork (74)
10:49:44.743 DRIVER     1 handler registered!
10:49:44.743 DRIVER     invoking handler #0
10:49:44.744 CNTRLR   handling add node request (status = NodeFound)
10:49:44.942 SERIAL « 0x0119004ad20315120421015e86725985737184803031707a5aef90            (27 bytes)
10:49:44.942 SERIAL » [ACK]                                                                   (0x06)
10:49:44.943 DRIVER « [REQ] [AddNodeToNetwork]
                        status:      AddingSlave
                        callback id: 210
10:49:44.944 DRIVER   handling request AddNodeToNetwork (74)
10:49:44.944 DRIVER     1 handler registered!
10:49:44.944 DRIVER     invoking handler #0
10:49:44.944 CNTRLR   handling add node request (status = AddingSlave)
10:49:44.949 DRIVER       the message was handled
10:49:45.440 SERIAL « 0x0107004ad205150070                                                 (9 bytes)
10:49:45.441 SERIAL » [ACK]                                                                   (0x06)
10:49:45.442 DRIVER « [REQ] [AddNodeToNetwork]
                        status:      ProtocolDone
                        callback id: 210
10:49:45.444 DRIVER   handling request AddNodeToNetwork (74)
10:49:45.444 DRIVER     1 handler registered!
10:49:45.444 DRIVER     invoking handler #0
10:49:45.444 CNTRLR   handling add node request (status = ProtocolDone)
10:49:45.444 CNTRLR   finishing inclusion process...
10:49:45.453 SERIAL » 0x0105004ac5d3a6                                                     (7 bytes)
10:49:45.454 DRIVER » [REQ] [AddNodeToNetwork]
                        action:       Stop
                        high power:   true
                        network wide: true
                        callback id:  211
10:49:45.460 SERIAL « [ACK]                                                                   (0x06)
10:49:45.461 SERIAL « 0x0107004ad306150072                                                 (9 bytes)
10:49:45.461 SERIAL » [ACK]                                                                   (0x06)
10:49:45.461 DRIVER « [REQ] [AddNodeToNetwork]
                        status:      Done
                        callback id: 211
10:49:45.465 SERIAL » 0x0105004ac50075                                                     (7 bytes)
[...]

PR #662, opened 2022-06-21
Fix node found bug

raman325:
fixes https://github.com/zwave-js/zwave-js-server/issues/661

We shouldn't be trying to dump the node yet because the node is not operational. This was a copy paste error.

### DIFF
--- a/src/lib/forward.ts
+++ b/src/lib/forward.ts
@@ -50,7 +50,6 @@ export class EventForwarder {
         this.sendEvent(client, {
           source: "controller",
           event: "node found",
-          node: dumpNode(node, client.schemaVersion),
         })
       );
       this.setupNode(node);


raman325:
@AlCalzone what properties of a node are available at the time of this event aside from nodeId? I think we should still provide a skeleton of the node so that upstream systems can get the node ID.

AlCalzone:
Node ID, device class, potentially supported/controlled CCs.

raman325:
for now let's merge this as this will have widespread impact

raman325:
> Node ID, device class of the root endpoint, potentially supported/controlled CCs. Maybe changing the event param type to something which indicates this better than `ZWaveNode` would make sense.

I agree with this btw, because it's implied that those properties should be accessible

PR #665, opened 2022-06-21
Fix node found state dump

raman325:
How does this look @AlCalzone as a stop gap until the driver does something better?

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -294,6 +294,12 @@ export type NodeState =
   | NodeStateSchema14
   | NodeStateSchema15;
 
+interface FoundNodeStateSchema19 extends EndpointStateSchema0 {
+  deviceClass: DeviceClassState | null;
+}
+
+export type FoundNodeState = FoundNodeStateSchema19;
+
 function getNodeValues(node: ZWaveNode, schemaVersion: number): ValueState[] {
   if (!node.ready) {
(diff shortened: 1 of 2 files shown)


AlCalzone (changes requested):

src/lib/state.ts
```diff
@@ -578,6 +587,23 @@ export const dumpNode = (node: ZWaveNode, schemaVersion: number): NodeState => {
...
+  foundNode: ZWaveNode,
+  schemaVersion: number
+): FoundNodeState => {
+  const base: FoundNodeStateSchema19 = {
+    nodeId: foundNode.nodeId,
+    deviceClass: foundNode.deviceClass
+      ? dumpDeviceClass(foundNode.deviceClass)
+      : null,
+    status: foundNode.status,
+    commandClasses: Array.from(foundNode.getSupportedCCInstances(), (cc) =>
```
raman325, line 600: this is safe right? From the code it seems like the CC map is initialized so worst case this will be empty?

AlCalzone: This should be safe.

FWIW using `getSupportedCCInstances` is an unnecessary complicated way to get to the CC information, since it creates an instance of each CC, but there's nothing better exposed atm.

When v10 is released, you can use `node.getCCs()` instead, which just returns the ID and version/supported/controlled/secure info.

PR #708, opened 2022-08-10
Add support for zwave-js v10

raman325:
Changelog: https://github.com/zwave-js/node-zwave-js/releases/tag/v10.0.0-beta.8

TODO:
- figure out s2 inclusion callbacks with change to options

### DIFF
--- a/src/lib/controller/message_handler.ts
+++ b/src/lib/controller/message_handler.ts
@@ -100,9 +100,15 @@ export class ControllerMessageHandler {
         return { success };
       }
       case ControllerCommand.beginExclusion: {
-        const success = await driver.controller.beginExclusion(
-          message.unprovision
-        );
+        const success =
+          message.unprovision !== undefined
+            ? await driver.controller.beginExclusion(message.unprovision)
+            : await driver.controller.beginExclusion(
+                message.strategy !== undefined
+                  ? { strategy: message.strategy }
+                  : undefined
+              );
+
         return { success };
       }
       case ControllerCommand.stopExclusion: {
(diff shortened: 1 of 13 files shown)


src/lib/node/message_handler.ts
```diff
@@ -211,6 +211,9 @@ export class NodeMessageHandler {
       case NodeCommand.waitForWakeup:
         await node.waitForWakeup();
         return {};
+      case NodeCommand.interview:
+        node.interview();
```
balloob, line 215: should this be awaited?

balloob (approved):

src/lib/node/message_handler.ts
raman325: The docs suggest not waiting for the interview to complete because of how long it can take

balloob: I think that we should await it. It's up to the caller (Home Assistant) to decide if they want to await the task or not.

raman325: should we do this in other places too? right now we don't await `node.interviewCC` for the same reason

balloob: I think so.

src/lib/controller/message_handler.ts
```diff
@@ -100,9 +100,15 @@ export class ControllerMessageHandler {
...
+            ? await driver.controller.beginExclusion(message.unprovision)
+            : await driver.controller.beginExclusion(
+                message.strategy !== undefined
+                  ? { strategy: message.strategy }
+                  : undefined
+              );
+
         return { success };
       }
       case ControllerCommand.stopExclusion: {
```
MartinHjelmare, line 114: I'm missing the changes needed for this point in the migration guide:

> - The Controller.replaceFailedNode method overload accepting the node ID as the second parameter has been removed. Use the overload with the ReplaceNodeOptions object instead. NOTE: If you do not pass this object, the new node will be included without security.

raman325: this was a deprecated signature that was removed. We switched to the new method a while ago, so there's no more that needs to be done here as far as I am aware.

PR #714, opened 2022-08-17
Bump zwave-js-server to 1.22.0

raman325:


MartinHjelmare:
I think we need another release that adds the rest of the changes needed for v10.

PR #739, opened 2022-09-09
Fix ZwavejsServer interface

raman325:
Both methods are async so shouldn't they be promises?

CC @robertsLando

### DIFF
--- a/src/lib/server.ts
+++ b/src/lib/server.ts
@@ -386,8 +386,8 @@ export interface Logger {
 }
 
 export interface ZwavejsServer {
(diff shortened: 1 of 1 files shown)


balloob (approved):

balloob:
No wait, how does this even work, it is exported under the same name? 🤔

balloob:
Anyway, to force keep it in sync, you can have the class implement the interface.

raman325:
I don't know, I just noticed that there were overloads for these functions. I will look into this again and open a new PR

PR #748, opened 2022-09-15
Bump zwave-js to 10.1.0 and add user agent component

raman325:
Changelog: https://github.com/zwave-js/node-zwave-js/releases/tag/v10.1.0

Here we update the user agent in the `ZwavejsServer` constructor call. This ensures that even when the server is embedded in another application like zwavejs2mqtt, the user agent includes the proper info.

### DIFF
--- a/src/lib/const.ts
+++ b/src/lib/const.ts
@@ -7,3 +7,4 @@ export const minSchemaVersion = 0;
 export const maxSchemaVersion = 22;
 
 export const applicationName = "zwave-js-server";
+export const dnssdServiceType = applicationName;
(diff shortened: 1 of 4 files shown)


src/lib/const.ts
```diff
@@ -6,4 +6,4 @@ export const minSchemaVersion = 0;
 // maximal/current schema version the server supports
 export const maxSchemaVersion = 22;
 
-export const dnssdServiceType = "zwave-js-server";
+export const applicationName = "zwave-js-server";
```
balloob, line 9: Let's create a new constant for this just so it's clear what we broadcast on DNS-SD

balloob (approved):

balloob (approved):

PR #749, opened 2022-09-15
Add user agent as parameter to controller.getAvailableFirmwareUpdates

raman325:
Implements https://github.com/zwave-js/node-zwave-js/pull/5070 which still needs to be released, hence draft


src/lib/controller/incoming_message.ts
```diff
@@ -243,6 +243,7 @@ export interface IncomingCommandControllerGetAvailableFirmwareUpdates
   command: ControllerCommand.getAvailableFirmwareUpdates;
   nodeId: number;
   apiKey?: string;
+  additionalUserAgentComponents?: Record<string, string>;
```
balloob, line 246: Wouldn't it make more sense to add this to the initial handshake? That way it is set once and can be used for other methods too.

balloob (approved):

PR #765, opened 2022-09-29
Add support for zwave-js 10.3

raman325:
We are mainly implementing https://github.com/zwave-js/node-zwave-js/pull/5121 and https://github.com/zwave-js/node-zwave-js/pull/5076

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,9 +458,9 @@ interface {
 }
 ```
 
-#### [Begin Firmware Update](https://zwave-js.github.io/node-zwave-js/#/api/node?id=beginfirmwareupdate)
+#### [Update Firmware](https://zwave-js.github.io/node-zwave-js/#/api/node?id=updatefirmware)
 
-[compatible with schema version: 5+]
+[compatible with schema version: 24+]
 
 If `firmwareFileFormat` is not provided, the format will be guessed based on the filename and file payload.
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -458,9 +458,9 @@ interface {
 }
 ```
 
-#### [Begin Firmware Update](https://zwave-js.github.io/node-zwave-js/#/api/node?id=beginfirmwareupdate)
+#### [Update Firmware](https://zwave-js.github.io/node-zwave-js/#/api/node?id=updatefirmware)
```
balloob, line 461: should we still docuemnt the old command?

src/lib/outgoing_message.ts
```diff
@@ -16,6 +16,7 @@ export type JSONValue =
   | number
   | boolean
   | null
+  | undefined
```
balloob, line 19: JSON does not have `undefined`. In fact, it's the greatest hack in JavaScript history, as undefined keys are removed from the output:

```
> JSON.stringify({hello: undefined})
"{}" 
```

raman325: yeah it was a typing issue with the result. I could use `as any` instead

README.md
raman325: I don't think so. It's deprecated and going to get removed in a future release. I left it for backwards compatibility but it should not be used by new clients

PR #772, opened 2022-10-04
Driver reset commands

### LINKED ISSUE #757: no way to reset zwave radio
I attempted to send command "driver.hard_reset" and got an unknown_command error code back.

```{"type":"result","success":false,"messageId":"66d9bed4-4c99-4046-8261-e1c2a00162cf","errorCode":"unknown_command","args":{"command":"driver.hard_reset"}}```

Let me know if there is any more information I can provide.

### PR DESCRIPTION
**Pull request by psittaciform**
Fixes #757 
Allows hard reset and soft reset to be called on the zwavejs driver from server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,8 @@ interface {
 
 #### [Instruct the controller to soft-reset (restart)](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
 
+[compatible with schema version: 25+]
+
 ```ts
 interface {
   messageId: string;
@@ -352,6 +354,8 @@ interface {
 
 #### [Instruct the controller to soft-reset (restart) only if soft-reset is enabled](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
 
+[compatible with schema version: 25+]
+
 ```ts
 interface {
   messageId: string;
@@ -360,6 +364,8 @@ interface {
 ```
 
 #### [Perform a hard reset on the controller](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=hardreset)
+
+[compatible with schema version: 25+]
 
 ```ts
 interface {
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ interface {
 }
 ```
 
-#### [Instruct the controller to soft-reset (restart)](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
+#### [Peform a soft-reset (restart) on the controller](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
 
 [compatible with schema version: 25+]
 
@@ -352,7 +352,7 @@ interface {
 }
 ```
 
-#### [Instruct the controller to soft-reset (restart) only if soft-reset is enabled](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
+#### [Attempt to perform a soft-reset (restart) on the controller](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
 
 [compatible with schema version: 25+]
 
--- a/README.md
+++ b/README.md
@@ -341,6 +341,33 @@ interface {
 }
 ```
 
+#### [Instruct the controller to soft-reset (restart)](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.soft_reset";
+}
+```
+
+#### [Instruct the controller to soft-reset (restart) only if soft-reset is enabled](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.try_soft_reset";
+}
+```
+
+#### [Perform a hard reset on the controller](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=hardreset)
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.hard_reset";
+}
+```
+
 ### Controller level commands
 
 `zwave-js-server` supports all of the controller methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/controller?id=controller-methods). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `controller.`, so `beginInclusion` is called using the `controller.begin_inclusion` command.
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -138,11 +138,15 @@ interface Args {
     try {
       // only start server if it does not already exist. Avoids port already in use error on driver reset.
       if (server == null) {
-        server = new ZwavejsServer(driver, {
-          port: args.port,
-          host: args.host,
-          enableDNSServiceDiscovery: !args["disable-dns-sd"],
-        });
+        server = new ZwavejsServer(
+          driver,
+          {
+            port: args.port,
+            host: args.host,
+            enableDNSServiceDiscovery: !args["disable-dns-sd"],
+          },
+          true
+        );
         await server.start();
       }
     } catch (error) {
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -136,7 +136,7 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      // don't start server if it already exists
+      // only start server if it does not already exist. Avoids port already in use error on driver reset.
       if (server == null) {
         server = new ZwavejsServer(driver, {
           port: args.port,
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -134,24 +134,17 @@ interface Args {
 
   let server: ZwavejsServer;
 
-  driver.on("driver ready", async () => {
-    try {
-      // only start server if it does not already exist. Avoids port already in use error on driver reset.
-      if (server == null) {
-        server = new ZwavejsServer(
-          driver,
-          {
-            port: args.port,
-            host: args.host,
-            enableDNSServiceDiscovery: !args["disable-dns-sd"],
-          },
-          true
-        );
-        await server.start();
-      }
-    } catch (error) {
-      console.error("Unable to start Server", error);
-    }
+  driver.once("driver ready", async () => {
+    server = new ZwavejsServer(
+      driver,
+      {
+        port: args.port,
+        host: args.host,
+        enableDNSServiceDiscovery: !args["disable-dns-sd"],
+      },
+      true
+    );
+    await server.start();
   });
 
   await driver.start();
--- a/src/bin/server.ts
+++ b/src/bin/server.ts
@@ -136,12 +136,15 @@ interface Args {
 
   driver.on("driver ready", async () => {
     try {
-      server = new ZwavejsServer(driver, {
-        port: args.port,
-        host: args.host,
-        enableDNSServiceDiscovery: !args["disable-dns-sd"],
-      });
-      await server.start();
+      // don't start server if it already exists
+      if (server == null) {
+        server = new ZwavejsServer(driver, {
+          port: args.port,
+          host: args.host,
+          enableDNSServiceDiscovery: !args["disable-dns-sd"],
+        });
+        await server.start();
+      }
     } catch (error) {
       console.error("Unable to start Server", error);
     }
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Review by raman325 — CHANGES_REQUESTED** [review]
Please bump the schema to 24 in `lib/const.ts` and include the required schema number in the README for these commands like others

**Comment by psittaciform** [comment]
@raman325 Let me know if there is anything else needed for this PR, thanks!

**Review by raman325 — APPROVED** [review]

**Comment by raman325** [comment]
this looks good to me but I am waiting for a second opinion. Thanks for your patience!

File: src/bin/server.ts [hunk 1]
```diff
@@ -136,12 +136,15 @@ interface Args {
...
-      });
-      await server.start();
+      // only start server if it does not already exist. Avoids port already in use error on driver reset.
+      if (server == null) {
+        server = new ZwavejsServer(driver, {
+          port: args.port,
+          host: args.host,
+          enableDNSServiceDiscovery: !args["disable-dns-sd"],
+        });
+        await server.start();
```
balloob, line 146: Since a restart of the driver requires a full reinitalization of the Z-Wave state by the clients, I would assume that if a server already exists, we should tell it to reset too: purge any state + disconnect all clients.

balloob: For ease of use we could even consider to just quit the process when we trigger a reset and let it just be automatically restarted ?

src/lib/driver/message_handler.ts
```diff
@@ -65,6 +65,15 @@ export class DriverMessageHandler {
       case DriverCommand.enableErrorReporting:
         driver.enableErrorReporting();
         return {};
+      case DriverCommand.softReset:
+        await driver.softReset();
```
balloob, line 69: These commands should raise an error if we are currently doing firmware upgrades, to prevent us from bricking devices.

raman325: this is a great call 👍🏾

src/bin/server.ts
raman325: yes I would agree with that first point. We can quit the process but if the process can purge state and disconnect all clients, it's effectively the same thing right?

balloob: It is effectively the same thing if it is done 100% right :)

raman325: does this work in the case where the server is embedded in zwave-js-ui? zwave-js-ui will restart itself on a hard reset, but only when done directly in the UI. If a WS client initiates a hard reset, there's no process to exit in that case because zwave-js-ui is the parent process for the server instance.

balloob: If Z-Wave JS UI restarts on a hard reset, this call should propagate up and restart when embedded in Z-Wave JS UI, or else things break.

Maybe we should allow passing in a hard reset handler to the constructor. ZJS-UI does it's thing, and our process also just quits.

src/lib/driver/message_handler.ts
psittaciform: Are you talking about end devices or the controller?  For end devices the upgrade process is resilient enough to handle being interrupted. as for the controller, shouldn't that be handled by the underlying zwave-js driver?  If not I will have to defer to someone smarter than me to implement these changes and I will close out this PR for now.

src/bin/server.ts
raman325: Or we can just emit an `exit` event and add a handler in zwave-js-ui

src/lib/driver/message_handler.ts
raman325: the controller. We'd have to look at the underlying logic for `softReset` and `hardReset` to see if it protects against that, I don't know offhand

raman325: I don't see anything that indicates protection around this.

raman325:
Thanks for the contributions. I will try resubmitting with some of the discussed changes.

raman325:
reopening this, I will make the necessary modifications here

raman325:
@robertsLando FYI the server emits a `hard reset` event when a hard reset request comes from a client. I see that you restart the application when this is called in zwave-js-ui, you should do the same when this event is emitted

README.md
```diff
@@ -341,6 +341,39 @@ interface {
 }
 ```
 
+#### [Instruct the controller to soft-reset (restart)](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
```
balloob, line 344: Shouldn't this also be perform ?

README.md
```diff
@@ -341,6 +341,39 @@ interface {
...
+[compatible with schema version: 25+]
+
+```ts
+interface {
+  messageId: string;
+  command: "driver.soft_reset";
+}
+```
+
+#### [Instruct the controller to soft-reset (restart) only if soft-reset is enabled](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
```
balloob, line 355: Why do we nee this? Can't the client just catch the error and inspect it ?

src/lib/driver/message_handler.ts
```diff
@@ -65,6 +65,24 @@ export class DriverMessageHandler {
...
+        if (driver.controller.isAnyOTAFirmwareUpdateInProgress()) {
+          throw new UnableToResetError("soft");
+        }
+        await driver.softReset();
+        return {};
+      case DriverCommand.trySoftReset:
+        if (driver.controller.isAnyOTAFirmwareUpdateInProgress()) {
+          throw new UnableToResetError("soft");
+        }
+        await driver.trySoftReset();
```
balloob, line 78: Is this actually a driver function, why ?

raman325: yes to the first question, @AlCalzone for the second

README.md
raman325: we don't, you are right that we can catch the error in the soft reset command. Only reason to keep it would be to match the driver API as closely as possible

src/lib/driver/message_handler.ts
AlCalzone: Error reports are sent to Sentry, so this is a way to opt-in.

src/lib/driver/message_handler.ts
```diff
@@ -14,6 +14,17 @@ export class DriverMessageHandler {
...
+      command in
+        [
+          DriverCommand.hardReset,
+          DriverCommand.softReset,
+          DriverCommand.trySoftReset,
+        ] &&
+      driver.controller.isAnyOTAFirmwareUpdateInProgress()
+    ) {
+      throw new FirmwareUpdateInProgressError();
+    }
```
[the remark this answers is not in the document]
AlCalzone: I guess it should

README.md
```diff
@@ -341,6 +341,39 @@ interface {
 }
 ```
 
+#### [Peform a soft-reset (restart) on the controller](https://zwave-js.github.io/node-zwave-js/#/api/driver?id=softreset)
```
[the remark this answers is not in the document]
raman325: if this moves to the driver then there is no need, but we should probably document the errors in a separate PR

src/lib/driver/message_handler.ts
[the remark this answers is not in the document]
raman325: https://github.com/zwave-js/node-zwave-js/pull/5220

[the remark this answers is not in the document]
raman325: we still need this check because we don't want to destroy the server instance only to have the hard reset fail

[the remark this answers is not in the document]
balloob: Why not destroy the service instance afterwards ?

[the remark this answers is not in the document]
balloob: because otherwise you need to keep this in sync and that will go wrong.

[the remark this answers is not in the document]
raman325: done

src/lib/server.ts
```diff
@@ -432,10 +432,10 @@ export class ZwavejsServer extends EventEmitter {
     });
     this.sockets = new ClientsController(this.driver, this.logger);
     this.sockets.on("hard reset", async () => {
-      this.emit("hard reset");
-      await this.destroy();
       this.driver.once("driver ready", () => this.start());
-      this.driver.hardReset();
+      await this.driver.hardReset();
+      await this.destroy();
```
balloob, line 437: Emitting still works when you destroy right? Or are we going to get a wrong order of events for listeners ?

raman325: yes because we are destroying all the objects the server holds but keeping the server instance which is still an EventEmitter

src/lib/server.ts
```diff
@@ -434,7 +435,9 @@ export class ZwavejsServer extends EventEmitter {
     this.sockets.on("hard reset", async () => {
       this.driver.once("driver ready", () => this.start());
       await this.driver.hardReset();
-      await this.destroy();
+      if (this.destroyServerOnHardReset) {
```
raman325, line 438: added this logic because `zwave-js-ui` will destroy the server on its own as part of the restart so best to let it do its thing instead of destroying the server twice.

src/lib/server.ts
raman325: made some changes so that we aren't destroying and recreating the server multiple times as part of this.

src/lib/server.ts
```diff
@@ -411,24 +413,37 @@ export class ZwavejsServer extends EventEmitter {
...
     }
 
+    this.driver.updateUserAgent({ [applicationName]: version });
     this.server = createServer();
     this.wsServer = new ws.Server({
       server: this.server,
       perMessageDeflate: true,
     });
     this.sockets = new ClientsController(this.driver, this.logger);
+    this.sockets.on("hard reset", async () => {
```
balloob, line 435: this event should be called `perform hard reset`. All other events are named after something that happened, ie `driver ready`. But in this case the server needs to be reset.

_It also feel a little weird that this is using the event system._

src/lib/server.ts
```diff
@@ -411,24 +413,37 @@ export class ZwavejsServer extends EventEmitter {
...
+      this.driver.once("driver ready", () => {
+        if (this.destroyServerOnHardReset) {
+          this.start();
+        }
+      });
```
balloob, line 440:
```suggestion
      if (this.destroyServerOnHardReset) {
        this.driver.once("driver ready", () => {
          this.start();
        });
      }
```

src/lib/server.ts
raman325: that's fair. I could make it an async function on the server instance and pass the instance up to the handler. That would require keeping a reference to the server in the ClientController and Client, but then the driver would handle all of the logic. Thoughts?

balloob: I don't have the architecture in my head atm. Do we have an architecture diagram I can reference ?

raman325: we don't but I can try to create one later this week

raman325: does this help? generated from the `classdiagram-ts` VSC extension:
image

balloob: seems passing a reference to server to ClientController is ok.

src/bin/server.ts
```diff
@@ -136,12 +136,19 @@ interface Args {
 
   driver.on("driver ready", async () => {
```
balloob, line 137: Can we use `driver.once` instead of guarding if the event handler already run ?

src/lib/server.ts
```diff
@@ -408,27 +445,41 @@ export class ZwavejsServer extends EventEmitter {
...
+    this.driver.updateUserAgent({ [applicationName]: version });
     this.server = createServer();
     this.wsServer = new ws.Server({
       server: this.server,
       perMessageDeflate: true,
     });
-    this.sockets = new ClientsController(this.driver, this.logger);
+    if (!this.remoteController) {
+      throw new Error("Remote controller not initialized");
+    }
```
balloob, line 477: We always know that it's initialized, so we don't need this check?

src/lib/server.ts
```diff
@@ -408,27 +445,41 @@ export class ZwavejsServer extends EventEmitter {
   private defaultHost: string = "0.0.0.0";
   private responder?: Responder;
   private service?: CiaoService;
+  private remoteController?: ZwavejsRemoteController;
```
balloob, line 448: It's initialized in the constructor
```suggestion
  private remoteController: ZwavejsRemoteController;
```

src/lib/server.ts
```diff
@@ -399,6 +413,29 @@ export interface Logger {
   debug(message: string): void;
 }
 
+export class ZwavejsRemoteController extends EventEmitter {
```
balloob, line 416: Can we add a JS Doc why this class exists.

balloob: Also, since it controls the server, should it be called something like `ZWaveJSServerRemoteController` ?

balloob (approved):

Issue #782, opened 2022-10-16
[Bug]: Bad ws-server "serverHost" leads to crash that prevents fixing the problem


LordMike:
### Checklist

- [ ]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

Docker

### Z-Wave JS UI version

8.1.0

### ZwaveJS version

10.3.0

### Describe the bug

For reasons, I had a hostname written in my "Home Assistant" WS Server "Server Host" setting. This value was wrong and does not exist anymore, which leads to this issue:

```
2022-10-16 19:27:01.927 INFO Z-WAVE: Node 85: value added 85-135-0-80-3 => 0
2022-10-16 19:27:01.928 INFO Z-WAVE: Node 85: value added 85-135-0-80-4 => 0
2022-10-16 19:27:01.929 INFO Z-WAVE: Node 85: value added 85-135-0-80-5 => 0
2022-10-16 19:27:01.931 INFO Z-WAVE: Node 85: value added 85-135-0-value => undefined
2022-10-16 19:27:01.937 INFO Z-WAVE: Node 85 ready: Aeotec Ltd. - ZWA019 (Water Sensor 7 Pro)
2022-10-16T17:27:01.938Z CNTRLR   [Node 085] The node is ready to be used
2022-10-16 19:27:01.941 INFO Z-WAVE: Node 92 is alive
2022-10-16T17:27:01.942Z CNTRLR   [Node 092] The node is alive.
2022-10-16 19:27:01.959 INFO Z-WAVE: Node 92 ready: AEON Labs - ZW090 (Z‐Stick Gen5 USB Controller)
2022-10-16T17:27:01.960Z CNTRLR   [Node 092] The node is ready to be used
Error: getaddrinfo ENOTFOUND core-zwave-js
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26)
[19:27:03] WARNING: Z-Wave JS UI crashed, halting add-on
[19:27:03] INFO: Z-Wave JS UI stoped, restarting...
s6-rc: info: service legacy-services: stopping
[19:27:03] INFO: Z-Wave JS UI stoped, restarting...
s6-svwait: fatal: supervisor died
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service legacy-cont-init: stopping
[...]

PR #783, opened 2022-10-23
Catch errors from WS server

raman325:
hasn't been an issue but https://github.com/zwave-js/zwave-js-server/issues/782 made me realize we should catch it for e.g. zwave-js-ui to handle it gracefully

### DIFF
--- a/src/lib/server.ts
+++ b/src/lib/server.ts
@@ -437,8 +437,8 @@ export class ZwavejsServer extends EventEmitter {
 
     this.logger.debug(`Starting server on ${localEndpointString}`);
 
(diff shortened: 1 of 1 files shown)


balloob (approved):

balloob:
Will `this.onError` be able to differentiate between errors from either object?

raman325:
> Will `this.onError` be able to differentiate between errors from either object?

No, should we add a source arg to the event?

raman325:
source class name can be retrieved from `sourceClass.constructor.name`

robertsLando (approved):

PR #906, opened 2023-04-09
Bump zwave-js to 10.14.1

raman325:
Implements the following commands:
- https://zwave-js.github.io/node-zwave-js/#/api/node?id=manuallyidlenotificationvalue
- https://zwave-js.github.io/node-zwave-js/#/api/node?id=setdateandtime

### DIFF
--- a/README.md
+++ b/README.md
@@ -815,6 +815,34 @@ interface {
 }
 ```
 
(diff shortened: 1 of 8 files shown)


raman325:
If @AlCalzone implements an overload for `manuallyIdleNotificationValue` that accepts a `ValueID` instead, we will switch to that since it is significantly easier for clients to use IMO

balloob (approved):

PR #956, opened 2023-06-11
Add support for zwave-js v11 (schema 29)

raman325:
This looks bigger than it is because I did some code clean up while making these changes:

- Added `{}` around case statements so we can use local constants for the responses
- Renamed different state interfaces to be consistent (`<MODEL>StateSchema<SCHEMA_NUM>`) and also updated schema numbers to reflect when they changed
- Switched `setValue` responses for `node`, `broadcast_node`, and `multicast_group` to result instead of success boolean
- Switched `controller.beginOTAFirmwareUpdate`, `controller.firmwareUpdateOTA`, `controller.firmwareUpdateOTW`, `node.beginFirmwareUpdate`, and `node.updateFirmware` responses to result instead of success boolean
- removed `unprovision` option from `controller.beginExclusion` command
- removed `name` and `info` from `ConfigurationValueMetadataStateSchema29`
- Updated callbacks for `firmware update progress` and `firmware update finished` events with no impact on clients


src/lib/state.ts
```diff
@@ -422,20 +418,25 @@ export const dumpConfigurationMetadata = (
...
-    newMetadata.type = "string";
+  if (schemaVersion < 2 && base.type === "buffer") {
+    base.type = "string";
   }
 
-  return newMetadata;
+  if (schemaVersion < 29) {
+    base.name = metadata.label;
+    base.label = metadata.description;
+  }
```
OXINARF, line 435: Shouldn't it return base here?

src/lib/const.ts
```diff
@@ -4,7 +4,7 @@ export const version = require("../../package.json").version;
 export const minSchemaVersion = 0;
```
OXINARF, line 4: Maybe I don't understand something, but how can the server support a client with a schema version < 29 when the return types (and also command arguments) have changed?

src/lib/state.ts
raman325: good catch!

src/lib/const.ts
raman325: that's a good point, thanks for calling that out!

src/lib/controller/message_handler.ts
```diff
@@ -102,11 +103,20 @@ export class ControllerMessageHandler {
...
       case ControllerCommand.beginExclusion: {
-        const success = await driver.controller.beginExclusion(
-          message.strategy !== undefined
-            ? { strategy: message.strategy }
-            : undefined
-        );
+        let strategy: { strategy: ExclusionStrategy } | undefined;
+        if (message.strategy === undefined) {
+          strategy =
+            message.strategy !== undefined
```
OXINARF, line 109: Something seems wrong here.

AlCalzone: @raman325 you may want to adopt what I did in the other PR before I saw this one:
https://github.com/zwave-js/zwave-js-server/pull/962/files#diff-9684276caee5caac11fdd0dd76254e6d8e3b9ed0072fe606d58381409711a43fR435-R455

That change essentially restores the functionality that was now removed from the driver.

raman325: > Something seems wrong here.

This is what happens when I work on this while half asleep 🤦🏾 

> you may want to adopt what I did in the other PR before I saw this one

yes this is cleaner, thanks!

src/lib/state.ts
```diff
@@ -422,20 +418,26 @@ export const dumpConfigurationMetadata = (
...
-  if (schemaVersion < 2 && newMetadata.type === "buffer") {
-    newMetadata.type = "string";
+  if (schemaVersion < 2 && base.type === "buffer") {
+    base.type = "string";
   }
 
-  return newMetadata;
+  if (schemaVersion < 29) {
+    base.name = metadata.label;
+    base.label = metadata.description;
```
OXINARF, line 434:
```suggestion
    base.info = metadata.description;
```

src/lib/common.ts
```diff
@@ -25,17 +25,13 @@ export function setValueOutgoingMessage(
...
 }
 
-export type ControllerFirmwareUpdateResultType =
-  | { result: ControllerFirmwareUpdateResult } // schemaVersion >= 29
-  | { success: boolean }; // schemaVersion < 29
-
-export type NodeFirmwareUpdateResultType =
-  | { result: FirmwareUpdateResult } // schemaVersion >= 29
+export type FirmwareUpdateResultType =
+  | { result: ControllerFirmwareUpdateResult | FirmwareUpdateResult } // schemaVersion >= 29
```
OXINARF, line 29: I don't know what is usually done here, but this type means that any client for this server will need to go check what the driver actually returns for each command that has this return type (that's basically already done on documentation anyway, but it makes it harder if you are checking an older server version, as the docs aren't versioned).

raman325: yeah this isn't the best way to type this, I've been playing with other options but I haven't found a good alternative yet. With that being said, since all of the communication is happening over websocket, the typing is really for typechecking and code quality for the server

AlCalzone (approved):

AlCalzone:
v11 was just released

PR #962, opened 2023-06-14
feat: update to zwave-js v11

AlCalzone:
Fixing the breaking changes that must be handled on the zwave-js-server side when upgrading zwave-js to v11.

### DIFF
--- a/src/lib/broadcast_node/outgoing_message.ts
+++ b/src/lib/broadcast_node/outgoing_message.ts
@@ -1,8 +1,8 @@
-import { VirtualValueID } from "zwave-js";
+import { VirtualValueID, SetValueResult } from "zwave-js";
 import { BroadcastNodeCommand } from "./command";
 
 export interface BroadcastNodeResultTypes {
-  [BroadcastNodeCommand.setValue]: { success: boolean };
+  [BroadcastNodeCommand.setValue]: { result: SetValueResult };
   [BroadcastNodeCommand.getEndpointCount]: { count: number };
   [BroadcastNodeCommand.supportsCC]: { supported: boolean };
   [BroadcastNodeCommand.getCCVersion]: { version: number };
--- a/src/lib/controller/incoming_message.ts
+++ b/src/lib/controller/incoming_message.ts
@@ -1,5 +1,6 @@
 import {
   AssociationAddress,
+  ExclusionOptions,
   ExclusionStrategy,
   FirmwareFileFormat,
   FirmwareUpdateFileInfo,
@@ -35,11 +36,18 @@ export interface IncomingCommandControllerStopInclusion
   command: ControllerCommand.stopInclusion;
 }
 
+// Schema >= ???
 export interface IncomingCommandControllerBeginExclusion
   extends IncomingCommandControllerBase {
   command: ControllerCommand.beginExclusion;
+  options?: ExclusionOptions;
+}
+
+// Schema <= ???
+export interface IncomingCommandControllerBeginExclusionLegacy
+  extends IncomingCommandControllerBase {
+  command: ControllerCommand.beginExclusion;
   unprovision?: boolean | "inactive";
-  strategy?: ExclusionStrategy;
 }
 
 export interface IncomingCommandControllerStopExclusion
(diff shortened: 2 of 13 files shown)


src/lib/broadcast_node/outgoing_message.ts
```diff
@@ -1,8 +1,8 @@
-import { VirtualValueID } from "zwave-js";
+import { VirtualValueID, SetValueResult } from "zwave-js";
 import { BroadcastNodeCommand } from "./command";
 
 export interface BroadcastNodeResultTypes {
-  [BroadcastNodeCommand.setValue]: { success: boolean };
+  [BroadcastNodeCommand.setValue]: { result: SetValueResult };
```
AlCalzone, line 5: @raman325  how do you usually handle return type changes? Just like this or do we need to decide on the schema version?

src/lib/controller/incoming_message.ts
```diff
@@ -35,11 +36,18 @@ export interface IncomingCommandControllerStopInclusion
   command: ControllerCommand.stopInclusion;
 }
 
+// Schema >= ???
```
AlCalzone, line 39: TODO: determine which schema version this is

src/lib/controller/message_handler.ts
```diff
@@ -298,28 +296,29 @@ export class ControllerMessageHandler {
...
-        success = await driver.controller.firmwareUpdateOTA(message.nodeId, [
-          message.update,
-        ]);
-        return { success };
+        const result = await driver.controller.firmwareUpdateOTA(
+          message.nodeId,
+          [message.update]
+        );
+        return { result };
       }
```
AlCalzone, line 304: This command is weird, since it does something different than the name suggests. I assume it is left in because of backwards compatibility?

src/lib/controller/message_handler.ts
```diff
@@ -298,28 +296,29 @@ export class ControllerMessageHandler {
...
       case ControllerCommand.firmwareUpdateOTA: {
-        success = await driver.controller.firmwareUpdateOTA(
+        const result = await driver.controller.firmwareUpdateOTA(
           message.nodeId,
           message.updates
         );
-        return { success };
+        return { result };
```
AlCalzone, line 310: Return type changed. Can we simply do this?

src/lib/controller/message_handler.ts
```diff
@@ -298,28 +296,29 @@ export class ControllerMessageHandler {
...
-        success = await driver.controller.firmwareUpdateOTW(
+        const result = await driver.controller.firmwareUpdateOTW(
           extractFirmware(
             file,
             message.fileFormat ??
               guessFirmwareFileFormat(message.filename, file)
           ).data
         );
-        return { success };
+        return { result };
```
AlCalzone, line 321: dito

src/lib/node/outgoing_message.ts
```diff
@@ -28,9 +30,9 @@ export interface NodeResultTypes {
   [NodeCommand.refreshValues]: Record<string, never>;
   [NodeCommand.refreshCCValues]: Record<string, never>;
   [NodeCommand.ping]: { responded: boolean };
-  [NodeCommand.hasSecurityClass]: { hasSecurityClass: boolean | "unknown" };
+  [NodeCommand.hasSecurityClass]: { hasSecurityClass: MaybeNotKnown<boolean> };
   [NodeCommand.getHighestSecurityClass]: {
-    highestSecurityClass: SecurityClass | undefined;
+    highestSecurityClass: MaybeNotKnown<SecurityClass>;
```
AlCalzone, line 35: These are equivalent, I just want to clarify the meaning of `undefined` here

src/lib/state.ts
```diff
@@ -295,7 +293,7 @@ type NodeStateSchema8 = NodeStateSchema7;
 type NodeStateSchema9 = NodeStateSchema8;
 
 interface NodeStateSchema10 extends NodeStateSchema9 {
-  highestSecurityClass: SecurityClass | undefined;
+  highestSecurityClass: MaybeNotKnown<SecurityClass>;
```
AlCalzone, line 296: dito

src/lib/state.ts
```diff
@@ -422,8 +420,6 @@ export const dumpConfigurationMetadata = (
     unit: metadata.unit,
     valueSize: metadata.valueSize,
     format: metadata.format,
-    name: metadata.name,
-    info: metadata.info,
```
AlCalzone, line 426: These properties were never really working. Does this removal need versioning?

raman325: we don't know whether consuming applications are expecting this data int he payload for e.g. validation, so safest to keep it in

PR #972, opened 2023-06-28
Add setInclusionUserCallbacks method to server

raman325:
Implements the functionality needed to complete https://github.com/zwave-js/zwave-js-ui/pull/3150 . We also now set the callbacks when starting the server in standalone mode (e.g. not embedded in another application).

CC @robertsLando @AlCalzone @balloob does this make sense? We will need some upstream changes to handle these callbacks when they are not initiated by HA


raman325:
Question @AlCalzone , when we provide callbacks as part of the inclusion options, do the callbacks that are stored in the driver options also get called? If so, with our current logic, we would have duplicate callbacks called.

raman325:
separately is it OK that the promises we return in the callbacks are stored in the same place regardless of whether the callback is part of InclusionOptions or driver options? Put another way, is there ever a scenario where an inclusion can be triggered by another controller and a user can start an inclusion with the primary controller at the same time?

robertsLando (approved):

AlCalzone:
> when we provide callbacks as part of the inclusion options, do the callbacks that are stored in the driver options also get called

No. The callbacks passed to the `beginInclusion` method have priority.

> is there ever a scenario where an inclusion can be triggered by another controller and a user can start an inclusion with the primary controller at the same time?

Probably yes, but the user would have to initiate the inclusion at the primary first, and then while that is going on, include a node using the inclusion controller. Honestly, I think we shouldn't worry about this. If this becomes a problem, I'd rather have the driver block/fail the inclusion initiated on the inclusion controller.

src/lib/inclusion_user_callbacks.ts
```diff
@@ -0,0 +1,88 @@
...
+            });
+          }
+        });
+      }
+      return clientsController.validateDSKAndEnterPinPromise;
+    },
+    abort: (): void => {
+      // settle the promises to ensure finally is triggered for the cleanup.
+      clientsController.grantSecurityClassesPromise?.resolve(false);
+      clientsController.validateDSKAndEnterPinPromise?.resolve(false);
```
AlCalzone, line 70: I would remove those two lines. `abort` is called by the driver at a point it has timed out waiting for the other promises to resolve.

src/lib/inclusion_user_callbacks.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Client, ClientsController } from "./server";
+
+export const inclusionUserCallbacks = (
+  clientsController: ClientsController,
+  client?: Client
+): InclusionUserCallbacks => {
+  return {
+    grantSecurityClasses: (
+      requested: InclusionGrant
+    ): Promise<InclusionGrant | false> => {
```
AlCalzone, line 12: Since you specify the return type of the surrounding method as `InclusionUserCallbacks`, the function types should entirely be inferred by TypeScript.
If anything, manually typing them again could introduce conflicts.

```suggestion
    grantSecurityClasses: (requested) => {
```

src/lib/inclusion_user_callbacks.ts
```diff
@@ -0,0 +1,88 @@
...
+              event: "grant security classes",
+              requested: requested as any,
+            });
+          }
+        });
+      }
+
+      return clientsController.grantSecurityClassesPromise;
+    },
+    validateDSKAndEnterPIN: (dsk: string): Promise<string | false> => {
```
AlCalzone, line 40:
```suggestion
    validateDSKAndEnterPIN: (dsk) => {
```
dito

src/lib/inclusion_user_callbacks.ts
```diff
@@ -0,0 +1,88 @@
...
+              source: "controller",
+              event: "validate dsk and enter pin",
+              dsk,
+            });
+          }
+        });
+      }
+      return clientsController.validateDSKAndEnterPinPromise;
+    },
+    abort: (): void => {
```
AlCalzone, line 67:
```suggestion
    abort: () => {
```
dito

AlCalzone (changes requested):

src/lib/inclusion_user_callbacks.ts
raman325: so what happens to the promises?

raman325: I guess we discard them and then the commands will throw an error if a client tries to complete the flow after the timeout

raman325: OK so I need to replace resolving them with deleting them. Will do that when i get home

PR #988, opened 2023-07-20
Bump zwave-js to 11.5.3 and schema to 30

raman325:
Adds lastSeen property to node dump and cleans up some code

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -269,11 +269,9 @@ type NodeStateSchema1 = Modify<
   { deviceClass: DeviceClassState | null; commandClasses: CommandClassState[] }
 >;
 
-type NodeStateSchema2 = NodeStateSchema1;
-
 type NodeStateSchema3 = Omit<
   Modify<
-    NodeStateSchema2,
+    NodeStateSchema1,
     {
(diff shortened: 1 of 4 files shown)


src/lib/state.ts
```diff
@@ -637,7 +636,13 @@ export const dumpNode = (node: ZWaveNode, schemaVersion: number): NodeState => {
     return node14;
   }
 
-  return node14 as NodeStateSchema15;
+  if (schemaVersion <= 29) {
+    return node14 as NodeStateSchema15;
+  }
+
+  const node30 = node14 as Partial<NodeStateSchema30>;
+  node30.lastSeen = node.lastSeen;
```
AlCalzone, line 644: Are you using the `as Partial` to stop TS from complaining that `lastSeen` is missing?
I would probably do this instead:
```ts
const node30: NodeStateSchema30 = {
    ...node14,
    lastSeen: node.lastSeen
}
```

raman325: that's fair, but can we stick with the original pattern for this PR and work on a better one in a separate PR?

AlCalzone (approved):

PR #997, opened 2023-07-24
ci: update outdated actions, cache dependencies

AlCalzone:
Fixes this warning on CI builds:
> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-node@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


raman325 (approved):
Thanks!

PR #1009, opened 2023-08-08
Bump zwave-js to 11.9.3 and add new node commands

raman325:
3 new node functions:
- `node.get_date_and_time`
- `node.is_health_check_in_progress`
- `node.abort_health_check`

And updated the callback from `node.check_lifeline_health` to include the last result

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "semver": "^7.5.4",
     "ts-node": "^10.9.1",
     "typescript": "^5.1.6",
-    "zwave-js": "^11.9.1"
+    "zwave-js": "^11.9.2"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "ws": "^8.0.0"
   },
   "peerDependencies": {
-    "zwave-js": "^11.5.3"
+    "zwave-js": "^11.9.1"
```
MartinHjelmare, line 37:
```suggestion
    "zwave-js": "^11.9.2"
```

AlCalzone (approved):

AlCalzone (approved):

MartinHjelmare:
The PR title should have been updated so the commit message matches the change.

raman325:
Agreed, missed that and thanks for calling it out. I'll see if I can amend the commit to master, I don't believe we have protection set up on the master branch so it should be doable

PR #1017, opened 2023-08-19
Deprecate listen to logs cmds from driver and add them in server

raman325:
Per the conversation in https://github.com/zwave-js/ZWaveJS.NET/pull/30

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 16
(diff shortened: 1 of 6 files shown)


AlCalzone:
I'm a bit confused. The PR titles says "move", but I only see added commands.

raman325:
> I'm a bit confused. The PR titles says "move", but I only see added commands.

for backwards compatibility we keep the old commands where they are. But I missed removing the old commands from the README because they're deprecated

AlCalzone (approved):

PR #1036, opened 2023-09-07
Add support for zwave-js 12 changes

raman325:
~~Still need to work out how to handle firmware updates~~

- Renamed methods/properties/events from heal* to rebuild routes
- Updated firmware updates for the new parameters. Clients that use the old parameters will get an error that they need to upgrade
- added endpoint to `notification` event
- for older clients, we will now send Z-Wave errors instead of regular errors for regular errors. This is so that we can pass the message which was being stripped before. For newer clients, we've added the message parameter to the regular error so we can continue to send them. For newer clients, we also now send the name of the Z-Wave error

TODO:
- [x] Clean up error logic (don't check for schema version in two places, make it consistent)


src/lib/driver/message_handler.ts
```diff
@@ -83,7 +82,8 @@ export class DriverMessageHandler {
         return {};
       }
       case DriverCommand.enableErrorReporting: {
-        driver.enableErrorReporting();
+        // This capability no longer exists but we keep the command here for backwards
```
raman325, line 85: should we throw an error here to match what we are doing with firmware updates? I think it's probably better to so that we are clear to the client what's happening

src/lib/controller/command.ts
```diff
@@ -5,9 +5,18 @@ export enum ControllerCommand {
   stopExclusion = "controller.stop_exclusion",
   removeFailedNode = "controller.remove_failed_node",
   replaceFailedNode = "controller.replace_failed_node",
+  // Schema version < 32
   healNode = "controller.heal_node",
+  // Schema version > 31
```
AlCalzone, line 10: Those version comparisons had me confused at first. They are correct, but they are counter-intuitive

src/lib/controller/message_handler.ts
```diff
@@ -132,16 +133,40 @@ export class ControllerMessageHandler {
...
+      // Schema < 32
       case ControllerCommand.healNode: {
-        const success = await driver.controller.healNode(message.nodeId);
+        const success = await driver.controller.rebuildNodeRoutes(
+          message.nodeId,
+        );
+        return { success };
+      }
+      // Schema >= 32
+      case ControllerCommand.rebuildNodeRoutes: {
```
AlCalzone, line 144: These two cases could be merged, but I guess you'll want to keep them separate in order to be able to add more logic to the 2nd one in the future?

src/lib/error.ts
AlCalzone, line 1: JavaScript errors are weird, since you cannot simply inherit from them. To make sure that stuff like `instanceof Error` works, you'll have to set the prototype manually in the constructor:
https://github.com/zwave-js/node-zwave-js/blob/d070781081bd2a81168efddb2822c29a36c7898e/packages/core/src/error/ZWaveError.ts#L230-L232

src/lib/error.ts
```diff
@@ -86,3 +87,14 @@ export class InclusionAlreadyInProgressError extends BaseError {
...
   errorCode = ErrorCode.invalidParamsPassedToCommand;
 }
+
+export class NoLongerSupportedError extends BaseError {
+  errorCode = ErrorCode.noLongerSupported;
+
+  constructor(public customMessage: string) {
+    super(
+      customMessage +
+        " If you are using an application that integrates with Z-Wave JS and you receive this error, you may need to update the application.",
```
AlCalzone, line 97: is the `customMessage` property actually used anywhere? If not, I wouldn't create another property for it:
```suggestion
  constructor(message: string) {
    super(
      message +
        " If you are using an application that integrates with Z-Wave JS and you receive this error, you may need to update the application.",
```
After this, the `error.message` property will be the full, concatenated string.

src/lib/forward.ts
```diff
@@ -333,13 +363,24 @@ export class EventForwarder {
...
-              args,
-            });
+            if (client.schemaVersion < 32) {
+              this.sendEvent(client, {
+                source: "node",
+                event: "notification",
+                nodeId: changedNode.nodeId,
+                ccId,
+                args,
+              });
```
AlCalzone, line 373: Do the clients have validation for extra properties? If not, we could simply pass the endpoint index independent of the schema version.

src/lib/server.ts
```diff
@@ -210,15 +215,27 @@ export class Client {
       if (err instanceof BaseError) {
         this.logger.error("Message error", err);
         const { errorCode, name, message, stack, ...args } = err;
-        return this.sendResultError(msg.messageId, errorCode, args);
+        if (this.schemaVersion < 32) {
+          this.sendResultZWaveError(
+            msg.messageId,
+            -1 as any,
```
AlCalzone, line 221: I suppose we don't expect this to be hit? Maybe add a comment to explain this?

src/lib/server.ts
```diff
@@ -210,15 +215,27 @@ export class Client {
...
+          );
+        }
+        return this.sendResultError(msg.messageId, errorCode, message, args);
       }
       if (err instanceof ZWaveError) {
         this.logger.error("Z-Wave error", err);
         return this.sendResultZWaveError(msg.messageId, err.code, err.message);
       }
 
       this.logger.error("Unexpected error", err as Error);
```
AlCalzone, line 232: To be 100% correct, err could be anything, not just an error. We can avoid the `as Error` by checking explicitly:
```ts
if (!(err instanceof Error)) {
    err = new Error(`Unexpected error: ${err}`);
}
```

src/lib/server.ts
```diff
@@ -210,15 +215,27 @@ export class Client {
...
-      this.sendResultError(msg.messageId, ErrorCode.unknownError, {});
+      this.sendResultError(
+        msg.messageId,
+        ErrorCode.unknownError,
+        (err as Error).toString(),
```
AlCalzone, line 236: Either
```suggestion
        err.message,
```
or
```suggestion
        err.stack,
```
depending on the level of detail you want to include.

AlCalzone (changes requested):

src/lib/controller/message_handler.ts
raman325: I'll combine them for now, we can always split them out later if/when needed

src/lib/forward.ts
raman325: they may. We had originally intended to do that for HA but couldn't with the validator we were using

src/lib/server.ts
raman325: no it could be hit, but what we are doing is finding a way to pass the message out because we didn't have that as an attribute before. So we are reusing the ZWaveError, and using `-1` so it doesn't overlap with any valid ZwaveErrorCodes

src/lib/error.ts
```diff
@@ -9,11 +9,19 @@ export enum ErrorCode {
...
 
 export class BaseError extends Error {
   // @ts-ignore
   errorCode: ErrorCode;
+
+  constructor(message?: string) {
+    super(message);
+    // We need to set the prototype explicitly
+    Object.setPrototypeOf(this, BaseError.prototype);
+    Object.getPrototypeOf(this).name = "ZWaveError";
```
AlCalzone, line 23:
```suggestion
    Object.getPrototypeOf(this).name = "BaseError";
```
but this would require overriding it in the subclasses too. If you don't care about having the error class name in stack traces, e.g.
```
NoLongerSupportedError: This is an error message
  at foo.js:80
  at bar.js:81
```
then we can just omit that line.

AlCalzone (changes requested):
Just one more thing

src/lib/error.ts
```diff
@@ -29,6 +29,9 @@ export class UnknownError extends BaseError {
 
   constructor(public error: Error) {
     super();
+    // We need to set the prototype explicitly
+    Object.setPrototypeOf(this, UnknownError.prototype);
```
AlCalzone, line 33: I don't think setting the prototype is necessary for subclasses, only the one directly inheriting from `Error`. I might be mistaken though.

raman325: From https://github.com/Microsoft/TypeScript-wiki/blob/main/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work 
> As a recommendation, you can manually adjust the prototype immediately after any `super(...)` calls.

> However, any subclass of `FooError` will have to manually set the prototype as well.

I'm interpreting that as meaning we **do** need to set the prototype of the subclasses, what do you think?

raman325:
Note that I switched from throwing an error to logging for `driver.enableErrorReporting` as this gets called automatically when opted in to data collection and starting the integration in HA

src/lib/error.ts
AlCalzone: Agreed. Reading it again, this may not be necessary depending on the emit target, but lets keep it to be safe.

PR #1068, opened 2023-10-15
Bump schema and update setRawConfigParameterValue cmd

raman325:
Looking at the types we didn't set this command up properly so updating it accordingly. Also adding it as an endpoint level command.

### DIFF
--- a/src/lib/common.ts
+++ b/src/lib/common.ts
@@ -54,9 +54,8 @@ export async function setRawConfigParameterValue(
   nodeOrEndpoint: ZWaveNode | Endpoint,
 ): Promise<{ result?: SupervisionResult }> {
   if (
-    message.valueSize !== undefined
-      ? message.valueFormat === undefined
-      : message.valueFormat !== undefined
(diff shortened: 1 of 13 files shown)


src/lib/common.ts
```diff
@@ -39,3 +46,42 @@ export function firmwareUpdateOutgoingMessage<
...
+export async function setRawConfigParameterValue(
+  message:
+    | IncomingCommandNodeSetRawConfigParameterValue
+    | IncomingCommandEndpointSetRawConfigParameterValue,
+  nodeOrEndpoint: ZWaveNode | Endpoint,
+): Promise<{ result?: SupervisionResult }> {
+  if (
+    message.valueSize !== undefined
+      ? message.valueFormat === undefined
+      : message.valueFormat !== undefined
```
AlCalzone, line 59: Interesting construct. I'd probably use `(... && ...) || (... && ...)` to be more obvious what this does

AlCalzone (changes requested):

AlCalzone (approved):

PR #1071, opened 2023-10-17
Add support for driver presets

raman325:
in this change, we expect `driverPresets` to be a potential key that is in the config passed to the server with a list of presets.


AlCalzone (changes requested):
As discussed on Discord, I'd do this a bit differently

AlCalzone (approved):

PR #1141, opened 2024-01-22
Bump zwave-js to 12.5.0 b1 and add support for LR keys

raman325:
I don't believe any additional functionality is required in the server for long range support but will continue to review the docs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,10 +35,10 @@
         "semver": "^7.5.4",
         "ts-node": "^10.9.2",
         "typescript": "^5.3.3",
-        "zwave-js": "^12.4.1"
+        "zwave-js": "^12.5.0-beta.1"
       },
       "peerDependencies": {
-        "zwave-js": "^12.4.0"
+        "zwave-js": "^12.5.0-beta.1"
       }
     },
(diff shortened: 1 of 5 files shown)


AlCalzone:
There are a couple of things that are not available for LR nodes - managing routes, neighbor info, etc.
Not sure if the server should deal with that though.

AlCalzone (approved):

raman325:
what happens when attempting to call those methods on a LR node? Will the method throw an error?

robertsLando (commented):
Tx @raman325! Looks good to me, I had to disable all the neighbors/priority/return routes calls too, dunno if the driver handles the error, maybe that's something that should be handled only from UI (hide/show some informations based on the protocol)

raman325:
Not sure I agree. Yes, the UI needs to show and hide functionality based on the node properties but the driver should also (IMO) be smart enough to tell the caller when they are attempting to do something that is unsupported

AlCalzone:
I considered raising an error, but since most of the route management methods would only log something and return `false` when it didn't work, I chose to go that route for now.
It's not like you are going to break something, it just doesn't do anything. Might consider converting to an error in the next major.

PR #1221, opened 2024-06-11
Add version info to mDNS advertisement

raman325:
Not a critical PR, just something I noticed that could be better


frenck (approved):
LGTM 👍

PR #1231, opened 2024-06-29
Add utils.lookup_device

marcus-j-davies:
@AlCalzone , @raman325 

This PR exposes the `lookupDevice` method of the `ConfigManager` class.

I think this should  be added, for the following reason.

Often, after scanning a QR Code, front ends might want to display the Device Information, present in the config.

 - Label
 - Manufatcure
 - Description
 
Generally, specifics about the device, not usually found in the QR its self.
I also use this method, during a Smart Start operation, to advise the user the device is unknown, as the return is either  a `DeviceConfig` or `undefined`

Example:

Screenshot 2024-06-29 at 17 12 44

You could argue, why not include this by calling `lookupDevice` during QR Code parsing within the server?
But to allow better 1:1 mapping of the API - I didn't, feel free to suggest otherwise.

Now, I am not too familiar with this code base, so let me know if something is not right with my changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1272,6 +1272,8 @@ interface {
 
 `zwave-js-server` supports all of the utility methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/utils). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `utils.`, so `parseQRCodeString` is called using the `utils.parse_qr_code_string` command.
 
+> NOTE: While some string utility commands are made available in the server due to their availability within the driver, this functionality works best when implemented locally and should not be used over WebSocket for any practical purpose.
+
 ### Config manager commands
 
 `zwave-js-server` supports all of the config manager methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/config-manager). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `config_manager.`, so `lookupDevice` is called using the `config_manager.lookup_device` command.
--- a/README.md
+++ b/README.md
@@ -1272,6 +1272,10 @@ interface {
 
 `zwave-js-server` supports all of the utility methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/utils). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `utils.`, so `parseQRCodeString` is called using the `utils.parse_qr_code_string` command.
 
+### Config manager commands
+
+`zwave-js-server` supports all of the config manager methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/config-manager). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `config_manager.`, so `lookupDevice` is called using the `config_manager.lookup_device` command.
+
 ## Events
 
 ### `zwave-js` Events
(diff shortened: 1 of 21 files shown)


raman325:
I wonder if this should be in a separate `config` namespace? (so `config.lookup_device` instead of `utils.lookup_device`). I know it's a lot to add a new namespace just for a single command but this might be a good place to put other config related methods

raman325:
or maybe `config_manager`

AlCalzone:
I vote for `utils.`. The config manager isn't going to get any more extensions, so you're looking at 2 methods total (lookup manufacturer by ID and device by fingerprint).

raman325:
@marcus-j-davies spoke with Al about this. I added all of the APIs for utils and config manager to be consistent. Sorry for the delayed activity on this but we should be good to merge soon!

src/lib/config_manager/command.ts
```diff
@@ -0,0 +1,11 @@
+export enum ConfigManagerCommand {
+  loadManufacturers = "config_manager.load_manufacturers",
```
AlCalzone, line 2: Missing:
- `loadAll`
- `configVersion`

src/lib/utils/command.ts
```diff
@@ -1,4 +1,9 @@
 export enum UtilsCommand {
   parseQRCodeString = "utils.parse_qr_code_string",
   tryParseDSKFromQRCodeString = "utils.try_parse_dsk_from_qr_code_string",
+  num2hex = "utils.num2hex",
+  formatId = "utils.format_id",
+  buffer2hex = "utils.buffer2hex",
```
AlCalzone, line 6: IMO formatting numbers/buffers by round-tripping through a websocket seems like unnecessary overhead.

src/lib/utils/command.ts
```diff
@@ -1,4 +1,9 @@
...
+  getEnumMemberName = "utils.get_enum_member_name",
```
AlCalzone, line 7: I'm a bit torn on this one, as it's essentially ``enum[value] ?? `unknown (${value})` ``

src/lib/utils/command.ts
```diff
@@ -1,4 +1,9 @@
...
+  rssiToString = "utils.rssi_to_string",
```
AlCalzone, line 8: this one is okay, since it's Z-Wave specific.

AlCalzone (changes requested):

src/lib/utils/command.ts
AlCalzone: discussed offline, make a note recommending implementing those locally.

src/lib/utils/command.ts
AlCalzone: discussed offline, make a note recommending implementing those locally.

src/lib/config_manager/command.ts
raman325: missing from docs too then :)

AlCalzone: Whoops

README.md
```diff
@@ -1272,6 +1272,8 @@ interface {
 
 `zwave-js-server` supports all of the utility methods listed in the [Z-Wave JS documentation](https://zwave-js.github.io/node-zwave-js/#/api/utils). `zwave-js-server` uses [snake casing](https://en.wikipedia.org/wiki/Snake_case) for commands and prefixes every controller command with `utils.`, so `parseQRCodeString` is called using the `utils.parse_qr_code_string` command.
 
+> NOTE: While some string utility commands are made available in the server due to their availability within the driver, this functionality works best when implemented locally and should not be used over WebSocket for any practical purpose.
```
AlCalzone, line 1275:
```suggestion
> NOTE: While some string utility commands like `num2hex`, `buffer2hex`, etc. are made available in the server due to their availability within the driver, this functionality works best when implemented locally and should not be used over WebSocket for any practical purpose.
```

AlCalzone (changes requested):
Just a small clarification

AlCalzone (approved):

PR #1236, opened 2024-07-08
feat: support zwave-js v13

AlCalzone:
Migration guide here: https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v13

I hope I didn't forget anything, at least the server compiles.


raman325 (approved):
Looks good to me, thanks for the help!

Note that we still need to do some lib work to support these changes before we can roll out support in Home Assistant

Issue #1199, opened 2024-05-12
TODO: Add Zniffer support


raman325:
- Add startup parameter to specify Zniffer port
- Handle starting and stopping the Zniffer
- Forward Zniffer events
- Figure out what to do with saved captures
- Figure out how to model this in the lib

PR #1250, opened 2024-08-15
Add zniffer support

raman325:
Adds zniffer support. Doesn't match the API 1/1 but considering the steps as outlined in the docs seemed to be performed in sequence, it didn't seem to make sense to split them out. Open to alternate suggestions though.

Fixes https://github.com/zwave-js/zwave-js-server/issues/1199

CC @AlCalzone

### DIFF
--- a/API_SCHEMA.md
+++ b/API_SCHEMA.md
@@ -76,4 +76,3 @@ Base schema.
 
 - Added controller `inclusion state changed` event
 - Added `config_manager` commands
-- Refactored message handlers to be stateful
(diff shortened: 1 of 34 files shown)


AlCalzone (approved):
One thing that still bugs me with this approach is the inability to let a user set up the region before starting the capture. It's probably not very relevant outside of testing setups where devices from multiple regions may be transmitting though.

raman325:
is the way to solve that to split init and start? I didn't think there was a practical difference but easy enough to split in two if needed

AlCalzone (changes requested):
I'd also split `.stop` from `.destroy`, so the capture can be stopped and started again without having to re-initialize.
Other than that, no objections.

AlCalzone (approved):

Issue #1216, opened 2024-06-05
Do not destroy server, driver or websocket connection on `hardReset`


AlCalzone:
When performing hard resets from HA or while using HA (e.g. through an external message sent to the server), the UX is pretty terrible. Sometimes the integration does not immediately reconnect due to timing issues. Redirecting to the new controller device page also does not work reliably (I suppose for the same reason).

According to https://github.com/zwave-js/zwave-js-server/pull/772 this was done because Z-Wave JS UI does it, but there wasn't really a reason for that either. The driver instance handles `hardReset` just fine without recreating - all that needs to be done is clean up the old nodes and cached information about the controller.

As of https://github.com/zwave-js/zwave-js-ui/pull/3748, Z-UI no longer restarts anything on hard reset. zwave-js-server should follow suit.

PR #1252, opened 2024-08-15
Fix hard reset

raman325:
Fix https://github.com/zwave-js/zwave-js-server/issues/1216


AlCalzone (approved):
Looks good. It seems there is a followup on the HA side necessary, as the hard reset isn't reacted to at all now until I reload the integration.

PR #1279, opened 2024-09-19
Code cleanup and fix Zniffer event bug

raman325:
- Made all message handlers implement the `MessageHandler` interface
- Fixed the event type for a `zniffer` event


raman325:
If an interface can describe methods then yes that makes sense. I think I originally added this because it fit my python mental model

AlCalzone:
Yep, you just have to say the class `implements MrssageHandler`

PR #1319, opened 2024-11-08
feat: forward `driver ready` event

AlCalzone:


raman325:
we never added this because the normal way to run the server, we only start listening after driver is ready, but with zjsui that's not a safe assumption

raman325 (approved):

src/lib/forward.ts
```diff
@@ -35,6 +35,17 @@ export class EventForwarder {
       this.setupNode(node),
     );
 
+    // Bind to driver events
+    this.clientsController.driver.on("driver ready", () => {
+      // forward event to all connected clients, respecting schemaVersion it supports
```
raman325, line 40: a) this comment is not accurate with the code right now and
b) we should bump the schema and only send this when its >=40 so that we don't break older clients

raman325 (changes requested):
whoops missed this on the first pass

AlCalzone:
The reason we need this is because it indicates the successful **end** of the hard reset sequence.

raman325:
yeah I noticed that after, so we can add the check here

raman325 (approved):
not sure if a filter is less or more efficient but I believe this is the same pattern we use elsewhere

AlCalzone:
I don't think it matters for the collection sizes we expect to deal with here.

PR #1324, opened 2024-11-11
Support controller.cancelSecureBootstrapS2

MindFreeze:
Needed for https://github.com/zwave-js/certification-backlog/issues/34

### DIFF
--- a/API_SCHEMA.md
+++ b/API_SCHEMA.md
@@ -86,3 +86,4 @@ Base schema.
 # Schema 40
 
 - Added `endpoint.try_get_node` command
+- Added `controller.cancelSecureBootstrapS2` command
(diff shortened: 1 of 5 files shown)


MartinHjelmare (commented):
Please bump the schema version, and update the API docs and readme.

MartinHjelmare:
If we include this with the v14 support before it's released (schema 40) we don't need a schema bump here.

MindFreeze:
Yea, let's get it into 40. I added it to that schema. Don't think README needs anything as it is a straight proxy to the driver and these are not explicitly explained there as far as I can see

MartinHjelmare:
Dominic or Raman should review the code.

PR #1344, opened 2024-11-19
refactor: convert to a hybrid ESM/CommonJS package

AlCalzone:
After migrating `zwave-js` to a hybrid ESM/CommonJS package, I noticed that Z-Wave JS UI contains both CommonJS and ESM sources of the same files in the bundle. I believe this is because Z-Wave JS Server is still CommonJS.

This PR converts the package into an ESM/CommonJS hybrid, meaning it supports being imported as ESM, as well as `require`d from CommonJS.


AlCalzone:
Confirmed to be working when compiling the HA addon with this. Z-UI's bundle size is down to old levels aswell.

PR #1394, opened 2025-02-28
feat: support Z-Wave JS v15

AlCalzone:


AlCalzone (commented):
TODO: Also add driver commands for the newly moved/renamed methods.

raman325 (approved):
sorry for the delay but looks good!

PR #1420, opened 2025-04-29
feat: add `sdkVersion` to `NodeState`, bump schema to 42

AlCalzone:
We noticed that the node dump did not include this field, although the driver makes it available.


raman325 (approved):
this works, thank you, but two quick thoughts:
1. I don't remember why but at some point I endedf up creating types for NodeState for every schema version, even if it was a replica of the previous version. We should probably be consistent whether we remove those or add the ones in between 35 and 42. Not needed for this PR though
2. Generally speaking, the node state dumping logic is messy. Is there any way to clean it up? Not a major issue and can be done in a separate PR, but it would be nice to make that function more maintainable

PR #1437, opened 2025-09-22
Bump Z-Wave JS to 15.14.0

balloob:
Changelog https://github.com/zwave-js/zwave-js/releases/tag/v15.14.0


raman325 (approved):
LGTM, and now I can say I assisted with the ESPHome proxy work

PR #1544, opened 2026-04-10
feat: support new credential API in Z-Wave JS 15.23.4, bump schema to 48

AlCalzone:
This PR adds support for the unifying credential API added in Z-Wave JS 15.23.2, which abstracts away most of the discrepancies between User Code CC and User Credential CC using appropriate mappings. The public types are modeled after User Credential CC.


raman325 (commented):
is there any value in returning to the caller which CC is being proxied? Is there ever a scenario where an endpoint can support both CCs?

AlCalzone:
It is possible, but the device needs to be switched between the two.
If that's the case I think Z-Wave JS should switch the device to User Credential CC automatically because that also allows for better scheduling.

So right now I think the application shouldn't care about the CC being used.

raman325 (approved):
not a comment on this PR more about the upstream changes, and it has already shipped, but I find it odd to see these APIs next to the more generic endpoint ones. Have you considered namespacing these out? I realize it would be a breaking change but just thinking about API shape and manageability

AlCalzone:
> Have you considered namespacing these out?

I've identified more opportunities to do the same with other CCs, so I might do that either way. Namespacing events though is probably too awkward.

AlCalzone:
I've moved the accessControl APIs in Z-Wave JS 15.23.1 into their own class accessible through `endpoint.accessControl`. Need to determine how we want to expose it here.

raman325:
My first thought would be to have separate message handlers for these CCs. So in the endpoint module, create separate sub-modules for each abstract CC with its own handler, incoming message, outgoing message, etc. Then we take the same instance/message handler logic from the server and add it to endpoint message handler.

raman325:
Whether it's in this PR or in another, with that implementation we can probably pull some of the logic into reusable methods in both places because the handling is more or less the same, its just how the message handler part of the string gets pulled out that changes

AlCalzone:
I now created an endpoint `access_control` submodule that can serve as a template for future feature APIs like this.

src/lib/endpoint/message_handler.ts
```diff
@@ -66,6 +55,10 @@ export class EndpointMessageHandler implements MessageHandler {
       endpoint = node;
     }
 
+    if (isAccessControlMessage(message)) {
+      return handleEndpointAccessControlCommand(message, endpoint);
+    }
```
raman325: doesn't need to be in this PR, but thinking through how this will work down the line, should isAccessControlMessage be generic and return the command handler? So something like:

```ts
msgHandler = getAbstractCCMessageHandler(message) // returns `handleEndpointAccessControlCommand`
if (msgHandler != null) return msgHandler(message, endpoint);

switch (message.command) {
...
```

raman325: we should probably mirror what we are doing in the server for this: https://github.com/zwave-js/zwave-js-server/blob/29942fa284ef3f5f236d6bf7cf238aab77133169/src/lib/server.ts#L194-L200

raman325: or update the server. Point is that we should be consistent in how we parse the messages and find the right handler

PR #1566, opened 2026-07-05
feat: route User Code CC setValue calls through access control API, bump schema to 51

raman325:
## Summary

zwave-js ([zwave-js/zwave-js#8930](https://github.com/zwave-js/zwave-js/pull/8930), targeting `next`) made the unified Access Control API the source of truth for User Code CC: the `userCode`/`userIdStatus`/`adminCode` values became internal and their `setValue`/`pollValue` support was removed. Clients that manage user codes via the `node.set_value` path (e.g. existing Home Assistant installs) would silently break once the server ships a driver containing that change.

This PR routes those legacy `setValue` calls through the access control API on a best-effort basis, covering every path the old `UserCodeCC` SET_VALUE implementation supported:

| Legacy `setValue` call | Routed to |
| --- | --- |
| `userCode` (string) | `accessControl.setCredential(...)` |
| `userIdStatus` = `Available` | `accessControl.deleteCredential(...)` |
| `userIdStatus` = `Enabled` | `accessControl.setUser(userId, { active: true })` |
| `userIdStatus` = `Disabled` | `accessControl.setUser(userId, { active: false })` |
| `userIdStatus` = `Messaging` | `accessControl.setUser(userId, { active: true, userType: NonAccess })` |
| `adminCode` / `masterCode` (string) | `accessControl.setAdminCode(...)` |

Not routed: `keypadMode` still works through `setValue` upstream, and `userIdStatus` = `PassageMode` has no access control equivalent (documented as no longer supported).

Implementation notes:

- The credential type is derived from `getCredentialCapabilitiesCached()` (PIN code, or Password for locks that allow non-PIN characters) rather than hardcoded
- The endpoint is resolved from `valueId.endpoint`, and anything unroutable (other properties/values, non-numeric propertyKey, unsupported endpoint/type) falls back to `node.setValue` unchanged
- Results are converted back to `SetValueResult` (`SetCredentialResult`/`SetUserResult` via local converters, `setAdminCode`'s supervision result via `supervisionResultToSetValueResult`) so the response shape is unchanged for callers

The routing intentionally applies to **all** schema versions: the clients that need it are the ones that have not adopted the access control API, and the legacy behavior it would otherwise preserve no longer exists in zwave-js. Schema 51 documents the behavior change; clients targeting schema 51+ should use the `endpoint.access_control` commands directly.

## Changes

- `src/lib/node/message_handler.ts`: `trySetUserCodeValue()` shim + result converters
- `src/lib/const.ts`: `maxSchemaVersion` 50 → 51
- `API_SCHEMA.md`: Schema 51 entry

## Testing

- `prettier --check`, `tsc --noEmit`, and `eslint` pass
- `tsx src/test/integration.ts` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/node/message_handler.ts
+++ b/src/lib/node/message_handler.ts
@@ -2,7 +2,12 @@ import {
   Driver,
   LifelineHealthCheckResult,
   RouteHealthCheckResult,
+  SetCredentialResult,
+  SetValueResult,
+  SetValueStatus,
+  ZWaveNode,
 } from "zwave-js";
+import { UserCredentialType, UserIDStatus } from "@zwave-js/cc";
 import {
   CommandClasses,
   ConfigurationMetadata,
@@ -12,7 +17,10 @@ import { NodeNotFoundError, UnknownCommandError } from "../error.js";
 import { Client } from "../server.js";
 import { dumpConfigurationMetadata, dumpMetadata, dumpNode } from "../state.js";
 import { NodeCommand } from "./command.js";
-import { IncomingMessageNode } from "./incoming_message.js";
+import {
+  IncomingCommandNodeSetValue,
+  IncomingMessageNode,
+} from "./incoming_message.js";
 import { NodeResultTypes } from "./outgoing_message.js";
 import {
   firmwareUpdateOutgoingMessage,
@@ -42,11 +50,20 @@ export class NodeMessageHandler implements MessageHandler {
 
     switch (message.command) {
       case NodeCommand.setValue: {
-        const result = await node.setValue(
-          message.valueId,
-          message.value,
-          message.options,
-        );
+        // zwave-js removed `setValue` support for User Code CC user codes,
+        // so route them through the access control API to keep legacy
+        // clients working
+        let result =
+          message.valueId.commandClass === CommandClasses["User Code"]
+            ? await trySetUserCodeValue(node, message)
+            : undefined;
+        if (result === undefined) {
+          result = await node.setValue(
+            message.valueId,
+            message.value,
+            message.options,
+          );
+        }
         return setValueOutgoingMessage(result, this.client.schemaVersion);
       }
       case NodeCommand.refreshInfo: {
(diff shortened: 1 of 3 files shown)


src/lib/node/message_handler.ts
```diff
@@ -370,3 +393,156 @@ export class NodeMessageHandler implements MessageHandler {
...
+      default:
+        // Other statuses (e.g. PassageMode) have no access control equivalent
+        return undefined;
+    }
+  }
+
+  // Setting a user code or clearing one (userIdStatus = Available) maps to
+  // the user's single credential
+  const isClear = property === "userIdStatus";
+  if (!isClear && typeof value !== "string") {
```
AlCalzone: **[Minor] · Correctness**

The removed `UserCodeCC` `SET_VALUE` accepted `userCode` as `string | Uint8Array`; this shim only accepts a string, so a binary code fails this guard, returns `undefined`, and falls through to `node.setValue` — which now `throwUnsupportedProperty`s.

**Why:** binary user codes silently stop working after the schema bump.

**Fix:** `accessControl.setCredential` already accepts `string | Uint8Array`, so accept `Uint8Array` here and drop the `value as string` cast below.

src/lib/node/message_handler.ts
```diff
@@ -370,3 +393,156 @@ export class NodeMessageHandler implements MessageHandler {
...
+  }
+
+  if (
+    (property !== "userIdStatus" && property !== "userCode") ||
+    typeof propertyKey !== "number"
+  ) {
+    return undefined;
+  }
+
+  if (property === "userIdStatus" && value !== UserIDStatus.Available) {
```
AlCalzone, line 433: **[Minor] · Correctness**

Routing `userIdStatus` = `Enabled`/`Disabled`/`Messaging` to `accessControl.setUser` throws `"users and codes must be stored together"` (the User Code CC branch) when the slot has no cached code. The old `SET_VALUE` re-sent with the existing (possibly undefined) code rather than throwing.

**Why:** this branch throws while the sibling `userCode` and clear branches return a `SetValueResult`, so the handler now has a mixed failure surface for callers.

**Fix:** decide deliberately — either catch and convert to a `SetValueResult`, or accept the throw as intended (arguably cleaner) behavior.

src/lib/node/message_handler.ts
raman325: Fixed in 1b25068 — the shim now accepts `Uint8Array` user codes. Also restructured so the type guard directly dominates the `setCredential` call, which lets TypeScript narrow `value` naturally and removes the cast entirely.

src/lib/node/message_handler.ts
raman325: Went with catch-and-convert in 1b25068: the shim's purpose is keeping the legacy `node.set_value` surface intact, so all routed paths should return a `SetValueResult`. The `Argument_Invalid` error is converted to `{ status: InvalidValue, message }` — consistent with the sibling conversions, which already use `InvalidValue` + message for the empty-slot results. Other error codes (e.g. communication failures) still propagate.

src/lib/node/message_handler.ts
```diff
@@ -370,3 +387,93 @@ export class NodeMessageHandler implements MessageHandler {
...
+
+  const accessControl = node.getEndpoint(endpointIndex ?? 0)?.accessControl;
+  if (accessControl === undefined) {
+    return undefined;
+  }
+
+  // User Code CC devices support exactly one credential type: Password
+  // instead of PINCode when the device allows non-PIN characters
+  const { supportedCredentialTypes } =
+    accessControl.getCredentialCapabilitiesCached();
```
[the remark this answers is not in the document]
raman325: Checked the implementation — `getCredentialCapabilitiesCached()` doesn't throw. Every path reads cached values with `?? fallback` defaults, so before the interview completes it returns empty/zero capabilities. In that case the `.find()` over supported credential types returns `undefined` and the shim already falls back via the existing `credentialType === undefined` check.

src/lib/node/message_handler.ts
```diff
@@ -370,3 +387,93 @@ export class NodeMessageHandler implements MessageHandler {
...
+  // For User Code CC devices the credential slot mirrors the user ID
+  return convertSetCredentialResultToSetValueResult(
+    isClear
+      ? await accessControl.deleteCredential(credentialType, propertyKey)
+      : await accessControl.setCredential(
+          propertyKey,
+          credentialType,
+          propertyKey,
+          message.value as string,
+        ),
```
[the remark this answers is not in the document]
raman325: `deleteCredential` has two overloads: `(type, slot)` and `(userId | undefined, type, slot)` — see `AccessControlAPI` in zwave-js. The 2-arg call here matches the first overload, compiles, and deletes the intended credential. The access-control message handler uses the 3-arg form because its incoming message includes an explicit `userId` filter.

src/lib/node/message_handler.ts
```diff
@@ -2,7 +2,12 @@ import {
   Driver,
   LifelineHealthCheckResult,
   RouteHealthCheckResult,
+  SetCredentialResult,
```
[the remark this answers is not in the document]
raman325: The main `zwave-js` entry does `export * from "./Node.js"`, so `SetCredentialResult` is a stable public export, not an incidental re-export. This file already imports everything else it needs (including `SetUserResult`) from `"zwave-js"`, so keeping the import here is the more consistent choice within this file.

src/lib/node/message_handler.ts
AlCalzone: I think we're still missing a code path that calls `addUser` when setting a user code on a previously unoccupied slot. Without this, there is no way to add a new user/code using this API.

raman325: Fixed in 77de0ed — when `getUserCached()` returns `undefined` for the slot, the shim now routes through `addUser(userId, {}, { type, slot, data })` instead of `setCredential`, creating the user and credential together. The result prefers the credential outcome when present, falling back to the user outcome when the user add itself fails on U3C devices.

AlCalzone (approved):
LGTM. Do you have any way to test this?

raman325:
If I build everything locally I can try the old services in HA which will exercise this path. I can't test all the conditions but I can validate at least the set and delete get done properly. I'll look into this later today

raman325:
tested and successfully working (I had to send Websocket commands manually since, per my comment on Discord, HA services are going to break and there's nothing we can do about it)

PR #1592, opened 2026-09-15
fix: capture removed node state before queueing events

AlCalzone:
The event queue introduced in #1580 can delay `dumpNode()` until after the driver removes the node from its map. The snapshot then throws `ZW0211` and the server drops the `node removed` event. Home Assistant can remain on "Removing device" after exclusion.

This PR fixes that by eagerly capturing the event payload for select events.


src/test/node_removed.ts
AlCalzone, line 1: remove this

src/lib/forward.ts
```diff
@@ -207,30 +208,31 @@ export class EventForwarder {
...
-          { minSchemaVersion: 29 },
+              node: nodeState,
+              ...(schemaVersion <= 28
+                ? {
+                    replaced: [
+                      RemoveNodeReason.Replaced,
+                      RemoveNodeReason.ProxyReplaced,
+                    ].includes(reason),
+                  }
+                : { reason }),
```
AlCalzone: Keep the split calls - that is easier to read.

src/lib/server.ts
```diff
@@ -491,10 +497,19 @@ export class ClientsController extends EventEmitter {
 
     // Forward one event per event loop operation - in case the `event()` function does expensive work.
     for (const client of recipients) {
+      let clientEvent = event;
```
AlCalzone:
```suggestion
      let eventPayload = event;
```

src/lib/server.ts
```diff
@@ -478,6 +483,7 @@ export class ClientsController extends EventEmitter {
     options?: {
       minSchemaVersion?: number;
       maxSchemaVersion?: number;
+      prepareImmediately?: boolean;
```
AlCalzone: lets rename this to `lazy` with a default of `true`

src/lib/server.ts
AlCalzone: Renamed to `eventPayload` in 4170db0.

src/lib/forward.ts
AlCalzone: Restored the separate schema-filtered calls in 4170db0.

src/lib/server.ts
AlCalzone: Renamed to `lazy` with deferred behavior by default in 4170db0. Both removal calls use `lazy: false`.

src/test/node_removed.ts
AlCalzone: Removed the file and its integration-test wiring in 4170db0.

